PR #1, opened 2017-03-28
First PR

dsgoers:
Add basic structure for API.


app/models/translation.rb
```diff
@@ -0,0 +1,15 @@
...
+class Translation < ActiveRecord::Base
+  belongs_to :resource
+  belongs_to :language
+  has_many :custom_pages
+
+  def add_new_version
+    Translation.create(resource: resource, language: language, version: version + 1)
+  end
+
+  def self.latest_translation(resource_id, language_id)
```
Omicron7: This should probably be done as an activerecord scope.

app/models/translation.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class Translation < ActiveRecord::Base
+  belongs_to :resource
+  belongs_to :language
+  has_many :custom_pages
+
+  def add_new_version
```
Omicron7: `create_next_version` might be abetter name.

app/controllers/drafts_controller.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+require 'rest-client'
+require 'digest/md5'
+require 'aws-sdk-rails'
+require 's3_helper'
+
+class DraftsController < ApplicationController
```
Omicron7: Spencer forwarded over a good PDF entitled Growing Rails Applications in Practice.

I'd recommend reading that and attempting to apply the Beautiful Controllers and Strong Parameters sections to this controller.

app/controllers/application_controller.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class ApplicationController < ActionController::Base
-  protect_from_forgery with: :exception
+  protect_from_forgery with: :null_session # TODO: Change back to :exception.
```
soberstadt: This line can be `skip_before_action :verify_authenticity_token`

app/models/translation.rb
dsgoers: Not sure what you mean by that... do you mean it shouldn't have `self?`

Omicron7: See http://guides.rubyonrails.org/active_record_querying.html#scopes
```ruby
class Translation < ActiveRecord::Base
    scope :latest_for, -> (resource_id, language_id) {
        where(resource_id: resource_id, language_id: language_id).order(version: :desc)
    }
end
```
This would create a `Translation.latest_for(resource_id, language_id)` method.
@soberstadt Would you recommend a scope here?

soberstadt: I wouldn't make a scope for the purpose of looking up 1 record, a class level method make sense.
However, if you ever are doing `where().first`, that can be replaced by a `find_by`, so this query could be re-written: `order(version: :desc).find_by(resource_id: resource_id, language_id: language_id)`

dsgoers:
OK, I think I've addressed all your comments.

PR #2, opened 2017-03-31
Json api and stuff

dsgoers:
--JSON api
--authorization
--refactoring


app/models/translation.rb
```diff
@@ -39,6 +40,22 @@ def delete_draft!
...
+      custom_pages.each do |custom_page|
+        if resource_page.id == custom_page.page_id
+          translated_pages.push(custom_page)
+        else
+          translated_pages.push(resource_page)
+        end
+      end
+    end
+
+    translated_pages
```
soberstadt: I could re-write this:
```
resource.pages.map do |resource_page|
  custom_pages.map do |custom_page|
    next custom_page if resource_page.id == custom_page.page_id
    resource_page
  end
end.flatten
```
But I think that you end up with an array that is filled with a lot of duplicates. What are you trying to accomplish here?

app/serializers/custom_page_serializer.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class CustomPageSerializer < ActiveModel::Serializer
+  attributes :id, :page_id, :translation_id, :structure
```
soberstadt: You don't need to include foreign keys if you also define the relationships.

app/models/translation.rb
dsgoers: We need to return a list of translated pages.  A translation can have a custom version of the page structure, and if it does, we want to replace the original page with the custom one.  Right now there's no translation happening because the structure of the page hasn't been ironed out yet.  So this is basically a placeholder.

soberstadt: So maybe something like:
```
resource.pages.map do |resource_page|
  custom_pages.find_by(page_id: resource_page.id) || resource_page
end
```
Would that do what you are thinking?

dsgoers: Yea, that looks like a simpler version.  What's the difference between `map` and `each?`  I'm not sure which one it is when I click Command+B in RubyMine.

soberstadt: http://stackoverflow.com/a/5254192/879524

dsgoers:
thanks

PR #3, opened 2017-04-03
Add a test and some refactoring

dsgoers:


spec/models/auth_token_spec.rb
```diff
@@ -8,6 +8,11 @@
     assert(result == :created)
   end
 
+  it 'generates a value for the token' do
+    result = AuthToken.create(access_code: AccessCode.find(1))
+    assert(!result.nil?)
```
soberstadt: We prefer to use `expect` over `assert`, so this line should be `expect(result).to_not be nil`. I think that reads more naturally.

app/models/translation.rb
```diff
@@ -24,7 +24,7 @@ def download_translated_page(page_filename)
                                     dev_hash: AuthHelper.dev_hash, locale: language.abbreviation,
                                     source_file_name: page_filename, export_file_name: page_filename }
   rescue RestClient::ExceptionWithResponse => e
-    return e.response
+    raise e
```
soberstadt: Now this becomes an unnecessary rescue too.

spec/s3_helper_spec.rb
```diff
@@ -8,12 +8,12 @@
     push
 
     pages_dir = Dir.glob('pages/*')
-    assert(pages_dir.empty?)
+    expect(pages_dir.empty?).to be(true)
```
soberstadt: When learning rspec, one of the lessons is that you (usually) don't have to do `.to be(true)` or `.to be(false)`.
Instead `expect(pages_dir).to be_empty` (ruby will read `be_xxx` and try `xxx?`)

Read more here: https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers

spec/s3_helper_spec.rb
```diff
@@ -8,12 +8,12 @@
...
   end
 
   it 'deletes zip file' do
     push
-    assert(!File.exist?('version_1.zip'))
+    expect(File.exist?('version_1.zip')).to be(false)
```
soberstadt: `expect(File).to_not exist('version_1.zip')`

dsgoers:
tanks

PR #4, opened 2017-04-04
Add latest translations array to serializer

dsgoers:


soberstadt:
👎

app/serializers/resource_serializer.rb
```diff
@@ -1,10 +1,20 @@
...
 
   belongs_to :system
 
   has_many :translations
   has_many :pages
+
+  has_many :latest_translations
+
+  def latest_translations
+    Translation.find_by_sql("select T.* from translations T inner join (select language_id, resource_id, max(version) as
```
soberstadt: A few things here.
1. Are you sure you need raw SQL for this? It is unclear what exactly this does.
2. If you do need it, consider moving it to an instance method on the Resource class.
3. Are you sure rubocop is running on this? I'm pretty sure this is not the best way to do a block of text, so I am surprised you got this past it.

spec/serializers/resource_serializer_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+require 'rails_helper'
+require Rails.root.join('app', 'serializers', 'resource_serializer.rb')
+
+describe ResourceSerializer do
+  it 'returns latest published translation for each language' do
+    resource = Resource.find(2)
+    serializer = ResourceSerializer.new(resource)
+    serialization = ActiveModelSerializers::Adapter.create(serializer)
+
+    latest_translations = serialization.as_json[:data][:relationships][:'latest-translations'][:data]
```
soberstadt: I have found it helpful to write my specs in this format:
```
it 'is well written' do
  # the lines needed to prepare for the test
  setup

  # as few of lines as possible to actually run the test
  result = do_something

  # all of the expectations
  expect(result).to be_fancy
  expect(1).to be (0)
end
```
Yummmm... spec sandwich 🍔

app/controllers/drafts_controller.rb
```diff
@@ -32,14 +32,14 @@ def create_new_draft
...
-      render json: 'Draft already exists for this resource and language.', status: 400
+      render json: 'Draft already exists for this resource and language.', status: :bad_request
     else
       existing_translation.create_new_version
     end
   end
 
   def publish
-    load_translation.publish
+    load_translation.update_draft(params)
```
dsgoers: I whitelist in Translation

dsgoers: https://github.com/CruGlobal/godtools-api-v2/blob/latest_translations/app/models/translation.rb#L30

app/serializers/resource_serializer.rb
dsgoers: No, I'm not sure I need raw SQL.  The purpose is to return the translation with the latest version for each language.

dsgoers: Rubocop is definitely running on this file.

soberstadt: I see the challenge in the problem you are trying to solve, and apart from caching the 'latest' version or using Postgres's "DISTINCT ON" plugin, I can't think of a better way of doing this, and neither of those stand out as preferable over this. Could you move this into the `Resource` class so it may be tested there and add some code comments as to how it works?

PR #6, opened 2017-04-05
Download and save translated name/description at publish time

dsgoers:


db/migrate/20170324171408_initial_migration.rb
```diff
@@ -8,6 +8,7 @@ def change
 
     create_table :resources do |t|
       t.string :name, null: false
+      t.string :description
```
soberstadt: Can you move these changes into their own migration file? Typically you don't want to be editing migrations.

dsgoers: I realize that, I just haven't run any yet, so I figured it would be OK to update the migration.

soberstadt: It's mostly good practice. I sometimes edit migrations within a PR, but once they have been merged I open a new one.

PR #7, opened 2017-04-06
Attributes

dsgoers:


spec/controllers/attributes_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require Rails.root.join('spec', 'support', 'mock_auth_helper.rb')
+
+describe AttributesController do
```
soberstadt: Again, I don't approve of controller tests. I think you should be writing integration tests: http://guides.rubyonrails.org/testing.html#integration-testing

spec/controllers/attributes_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      attribute = Attribute.find(1)
+      expect(attribute.key).to eq('foo')
+      expect(attribute.value).to eq('new value')
+      expect(attribute.resource_id).to eq(1)
+    end
+
+    it 'deletes an Attribute' do
+      delete :destroy, params: { id: 1 }
+
+      expect { Attribute.find(1) }.to raise_error('Couldn\'t find Attribute with \'id\'=1')
```
soberstadt: You could just test `expect { Attribute.exists?(1) }.to be false`

soberstadt: Also these hard-coded integers will probably get you in trouble down the road. I would suggest something like `let(:attribute_id) { Attribute.first.id }` and using `attribute_id` throughout.

Reference: https://www.relishapp.com/rspec/rspec-core/v/2-5/docs/helper-methods/let-and-let

spec/controllers/attributes_controller_spec.rb
dsgoers: I don't see why I shouldn't test controllers.  They're part of the application, so they should be tested.  I don't want to write integration tests to test these really simple controllers...

soberstadt: Here are two places that give some explanation on why integration over controller:
http://betterspecs.org/#integration
http://stackoverflow.com/questions/6129694/should-i-really-test-controllers

I think that most of you controller tests could be converted to integration tests quickly. It seems like adding `type: :request` to the top `describe` and changing your request calls to something like `post 'attributes/create
, attribute: { key: 'foo', value: 'bar', resource_id: 1 } ` that should get you well on your way to make these tests include more of the stack. (I think that link on rails 5 integration tests was sent in haste, I didn't realize that was for minitest.)

spec/controllers/attributes_controller_spec.rb
dsgoers: I will add a constants file that has the IDs for all the seed objects, but I'll do it in another PR.  🍪 ?

spec/controllers/attributes_controller_spec.rb
dsgoers: I still disagree.  Controllers are part of the application and have methods that are called, so testing them is important.  Testing them helps verify they are thin and don't have too much logic.  Integration tests aren't what I want because I simply want to test my controllers call the right methods and pass the right values.

PR #9, opened 2017-04-10
Add .env file

dsgoers:


dsgoers:
Ok, ready for review.

config/database.yml
```diff
@@ -23,5 +23,12 @@ development:
 test:
   <<: *default
```
knutsenm: Having the same config (database name in particular) for development and test will cause you problems locally. See the comments here, and see other projects' database.yml for a proper setup.

.env.local
```diff
@@ -0,0 +1,9 @@
+DB_HOST=localhost
```
knutsenm: This file shouldn't be checked in. It may have snuck into your commit before you added it to .gitignore

PR #10, opened 2017-04-10
Use test constants so I'm not hardcoding IDs

dsgoers:


dsgoers:
Yea, I did this for readability

PR #13, opened 2017-04-13
Allow Resources to be filtered by System name

dsgoers:


app/controllers/resources_controller.rb
```diff
@@ -1,12 +1,24 @@
...
+  end
+
   def show
     resource
   end
 
   private
 
+  def all_resources
+    if params[:system]
```
frett: tossing this comment here so brian can suggest the correct syntax.

for this filtering option, can we name the parameter `filter[system]` to coincide with the jsonapi spec around filtering? see: http://jsonapi.org/format/#fetching-filtering

PR #14, opened 2017-04-18
Include translations

dsgoers:
1. Don't include any translations unless specified
2. Include translations with all resources


dsgoers:
thanks!!

PR #19, opened 2017-04-21
Add rollbar

dsgoers:


config/initializers/rollbar.rb
```diff
@@ -0,0 +1,56 @@
...
+  # Add exception class names to the exception_level_filters hash to
+  # change the level that exception is reported at. Note that if an exception
+  # has already been reported and logged the level will need to be changed
+  # via the rollbar interface.
+  # Valid levels: 'critical', 'error', 'warning', 'info', 'debug', 'ignore'
+  # 'ignore' will cause the exception to not be reported at all.
+  # config.exception_level_filters.merge!('MyCriticalException' => 'critical')
+  #
+  # You can also specify a callable, which will be called with the exception instance.
+  # config.exception_level_filters.merge!('MyCriticalException' => lambda { |e| 'critical' })
```
soberstadt: I would suggest adding `config.exception_level_filters.merge!('ActionController::RoutingError' => 'ignore')` because we usually get crawlers that hit random urls like 'wp-admin.php' on our servers to find variabilities.

PR #20, opened 2017-04-21
Add index endpoint to translations controller

ryancarlson:
To allow clients to download list of translations, filtered by is_published for now, and include other things that it cares about.

### DIFF
--- a/app/controllers/translations_controller.rb
+++ b/app/controllers/translations_controller.rb
@@ -1,12 +1,25 @@
 # frozen_string_literal: true
 
 class TranslationsController < ApplicationController
+  
+  def index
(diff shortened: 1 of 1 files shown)


app/controllers/translations_controller.rb
```diff
@@ -1,12 +1,24 @@
...
+
   def show
     redirect
   end
 
   private
 
+  def all_translations
+    if params['filter']
+      Translation.is_published(params['filter']['is_published'])
```
dsgoers: I'm guessing you based this off `Resource.system_name.`  You'll need to create a scope in `Translation` if you want this to work.  However, I don't think we want to return drafts here, since this is an unsecured endpoint.  This should probably only return drafts

PR #22, opened 2017-04-24
Use postgres config instead of mysql for doc generation

dsgoers:


dsgoers:
I tested this and it works.

PR #23, opened 2017-04-26
Add key so relationship name and include value are the same

dsgoers:


app/serializers/resource_serializer.rb
```diff
@@ -7,6 +7,6 @@ class ResourceSerializer < ActiveModel::Serializer
   belongs_to :system
 
   has_many :translations
-  has_many :latest_translations
+  has_many :latest_translations, key: 'latest-translations'
```
knutsenm: I'm not familiar w/that; is it an alias for primary_key or foreign_key? Should probably spell it one of those ways.

dsgoers:
thanks!!

PR #24, opened 2017-04-27
Insert translated phrases into XML structure documents

dsgoers:


app/controllers/translations_controller.rb
```diff
@@ -21,7 +21,8 @@ def all_translations
   end
 
   def render_not_found
-    @translation.errors.add(:message, 'Translation not found.')
+    @translation.errors.add(:message,
+                            'Translation not found.  Use drafts/ if you\'re looking for an unpublished translation.')
```
knutsenm: Rubocop won't complain if you use double quotes in this case (so you don\'t have to quote your single quote) ;-)

It would otherwise complain about double quotes when you don't interpolate any variables, so your intentions are noble.

dsgoers: I prefer to use single quotes and escape characters.

PR #26, opened 2017-04-27
Attributes

dsgoers:
1. Show resource attributes as attributes instead of a relationship
2. Force attribute keys to be in format blah_blah_blah

### DIFF
--- a/app/models/attribute.rb
+++ b/app/models/attribute.rb
@@ -4,4 +4,6 @@ class Attribute < ActiveRecord::Base
   belongs_to :resource
 
   has_many :translated_attributes
+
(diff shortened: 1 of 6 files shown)


app/models/attribute.rb
```diff
@@ -4,4 +4,6 @@ class Attribute < ActiveRecord::Base
   belongs_to :resource
 
   has_many :translated_attributes
+
+  validates :key, format: { with: /\A[[:alpha:]]+(_[[:alpha:]]+)*\z/ }
```
dsgoers: Yup that's it.  Might be too strict, but I can adjust it if needed.  Rails makes it easy 😄

PR #27, opened 2017-04-28
Json api

dsgoers:


app/controllers/custom_pages_controller.rb
```diff
@@ -13,11 +13,13 @@ def destroy
   private
 
   def upsert_custom_page
-    CustomPage.create!(params.permit(:translation_id, :page_id, :structure))
+    attrs = params[:data][:attributes]
+
+    CustomPage.create!(attrs.permit(:translation_id, :page_id, :structure))
```
soberstadt: You also can use the AMS deserializer to pull off the correct attributes, if you want.
https://github.com/CruGlobal/missionhub-api/blob/master/app/controllers/apis/v4/base_controller.rb#L202

dsgoers: What are the advantages/disadvantages?

soberstadt: I like it because it is more clear that what you are doing is parsing someone's json-api model, and it will do some normalizing of relationships. For example:
this hash:
```
{
  type: 'group',
  attributes: {
    name: 'The Avengers',
    location: 'Avengers Mansion',
    meets: 'weekly'
  },
  relationships: {
    organization: {
      data: {
        type: 'organization',
        id: org.id
      }
    }
  }
}
```
is parsed as such: `{:name=>"The Avengers", :location=>"Avengers Mansion", :meets=>"weekly", :organization_id=>4}`

Disadvantage is that it's a different api you need to remember.

I would recommend using it.

PR #28, opened 2017-04-28
Add strict mode for translating pages because sometimes we want to vi…

dsgoers:
…ew a page before we've translated all the phrases

### DIFF
--- a/app/controllers/drafts_controller.rb
+++ b/app/controllers/drafts_controller.rb
@@ -51,16 +51,14 @@ class DraftsController < SecureController
 
   def edit
     translation = load_translation
-    begin
-      translation.update_draft(params)
-      head :no_content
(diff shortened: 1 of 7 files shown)


app/controllers/drafts_controller.rb
```diff
@@ -47,7 +47,17 @@ def load_resource
   end
 
   def edit
-    load_translation.update_draft(params)
+    translation = load_translation
+    begin
```
soberstadt: You don't need to do a `begin/end`, instead:
```
def edit
  translation = load_translation
  translation.update_draft(params)
  head :no_content
rescue Error::PhraseNotFoundError => e
  translation.errors.add(:id, e.message)
  render json: translation,
           status: :conflict,
           adapter: :json_api,
           serializer: ActiveModel::Serializer::ErrorSerializer
end
```

PR #29, opened 2017-05-01
Onesky phrase

dsgoers:


app/models/translation_element.rb
```diff
@@ -2,4 +2,12 @@
 
 class TranslationElement < ActiveRecord::Base
   belongs_to :page
+
+  before_create :set_onesky_phrase_id
+
+  private
+
+  def set_onesky_phrase_id
+    self.onesky_phrase_id = SecureRandom.uuid unless onesky_phrase_id.present?
```
soberstadt: you also could do `self.onesky_phrase_id ||= SecureRandom.uuid`

app/models/auth_token.rb
```diff
@@ -3,10 +3,7 @@
 class AuthToken < ActiveRecord::Base
   belongs_to :access_code
 
-  validates :access_code, presence: true
```
soberstadt: While you may have DB constraints, I typically prefer to lean on application level validation. DB constraints is obfuscated rules. Having both makes it clear what will happen when you try to save.

dsgoers: Grrrrrr I know... but repeating...

PR #30, opened 2017-05-03
Validations

dsgoers:


spec/models/custom_page_spec.rb
```diff
@@ -5,9 +5,9 @@
 describe CustomPage do
   it 'cannot be replicated' do
     expect do
-      CustomPage.create(translation_id: 3,
-                        page_id: 1,
-                        structure: '{ <xml>updated structure</xml> }')
-    end.to raise_error(ActiveRecord::RecordNotUnique)
+      CustomPage.create!(translation_id: 3,
```
soberstadt: Another option for this is something like:
```
page = CustomPage.new(translation_id: 3, page_id: 1, structure: '{ <xml>updated structure</xml> }')

expect(page).to_not be_valid
```

dsgoers: I like testing the error message though... can you do so with `be_valid`?

soberstadt: You can do something like this: https://github.com/CruGlobal/missionhub-api/blob/3949862768860222b17ed01a41e865a88b5593bc/spec/models/v4/person/csv_generator_spec.rb#L121-L122

PR #33, opened 2017-05-04
Update onesky phrases

dsgoers:


app/controllers/resources_controller.rb
```diff
@@ -1,12 +1,23 @@
...
     render json: all_resources, include: params[:include], status: :ok
   end
 
   def show
-    resource
+    render json: load_resource, include: params[:include], status: :ok
+  end
+
+  def update
+    a = authorize!
```
Omicron7: This should probably be implemented using a filter. It'll allow you to enforce authorized! before this action is called.
See http://guides.rubyonrails.org/action_controller_overview.html#filters
Something like
```ruby
before_action :authorize!, only: :show
```
You can actually add it to a base controller to enforce all actions, the use `skip_before_action` in specific controllers to skip authorization if it isn't needed.

PR #36, opened 2017-05-09
Gt 277

dsgoers:


dsgoers:
Hold off on reviewing this for now.

dsgoers:
@Omicron7 OK, ready for review.

app/controllers/attachments_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class AttachmentsController < SecureController
+  def create
+    Attachment.create!(params.permit(permitted_params))
+    head :created
+  end
+
+  def update
+    load_attachment.update(params.permit(permitted_params))
```
Omicron7: If attachment isn't found, this will raise an exception, method update not found. Will this generate the correct jsonapi response?

app/controllers/attachments_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+    head :no_content
+  end
+
+  def destroy
+    load_attachment.destroy
```
Omicron7: Same comment as attachment update.

app/models/attachment.rb
```diff
@@ -0,0 +1,33 @@
...
+  private
+
+  def resource_or_translation
+    raise 'Attachment must be related to Resource or Translation.' if resource_id.nil? && translation_id.nil?
+
+    return unless resource_id.present? && translation_id.present?
+    raise 'Attachment can be related to Resource OR Translation, not both.'
+  end
+
+  def duplicate_keys # TODO: would be nice to enforce this in the DB
```
Omicron7: This might be possible with table redesign and activerecord inheritance.
https://apidock.com/rails/ActiveRecord/Inheritance

dsgoers:
Rebased against develop.

app/controllers/attachments_controller.rb
dsgoers: If attachment isn't found a 404 exception is raised.

Omicron7: And it renders it correctly for jsonapi?

dsgoers: I agree though that it should be JSON api friendly

app/models/attachment.rb
dsgoers: Yea, I just put the note there because I'm not sure I want to do it yet

app/controllers/attachments_controller.rb
dsgoers: Pushed up a commit for this.

PR #37, opened 2017-05-09
If we get a JSON parsing error, assume it's because there's no transl…

dsgoers:
…ated phrases available for the language and therefore no data was returned

### DIFF
--- a/app/models/translation.rb
+++ b/app/models/translation.rb
@@ -48,9 +48,9 @@ class Translation < ActiveRecord::Base
                               params: { api_key: ENV['ONESKY_API_KEY'], timestamp: AuthUtil.epoch_time_seconds,
                                         dev_hash: AuthUtil.dev_hash, locale: language.code,
                                         source_file_name: page_filename, export_file_name: page_filename }
-    return JSON.parse(response)
-  rescue JSON::ParserError
-    raise Error::PhraseNotFoundError, 'No translated phrases found for this language.'
+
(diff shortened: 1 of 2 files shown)


app/models/translation.rb
```diff
@@ -43,12 +43,14 @@ def build_translated_page(page_id, strict)
...
-                   params: { api_key: ENV['ONESKY_API_KEY'], timestamp: AuthUtil.epoch_time_seconds,
-                             dev_hash: AuthUtil.dev_hash, locale: language.code,
-                             source_file_name: page_filename, export_file_name: page_filename }
+    response = RestClient.get "https://platform.api.onesky.io/1/projects/#{resource.onesky_project_id}/translations",
+                              params: { api_key: ENV['ONESKY_API_KEY'], timestamp: AuthUtil.epoch_time_seconds,
+                                        dev_hash: AuthUtil.dev_hash, locale: language.code,
+                                        source_file_name: page_filename, export_file_name: page_filename }
+    return JSON.parse(response)
+  rescue JSON::ParserError
+    raise Error::PhraseNotFoundError, 'No translated phrases found for this language.'
```
soberstadt: This seems like it would be likely to catch other errors that you might be throwing away. Like timeouts or auth errors. Could you check the status code instead?

dsgoers: Which status do you mean?  The HTTP status wouldn't be of any help here, if that's what you mean.

PR #38, opened 2017-05-10
provide resource dereferencing via the manifest xml

frett:
This will allow page xml to be a bit more readable and not require rewriting resources when publishing.
This also provides a quick list of all resources used within a tool.

### DIFF
--- a/manifest/content.xml
+++ b/manifest/content.xml
@@ -13,10 +13,10 @@
 
   <!--
     Image Content - WIP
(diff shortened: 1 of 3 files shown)


frett:
I figure this is a good mechanism to discuss any spec changes and make each other aware of them. @ryancarlson @pablomarti

frett:
bumping this while @ryancarlson and @pablomarti are already looking at content spec PRs ;)

PR #39, opened 2017-05-10
Changes to languages and drafts endpoints

dsgoers:


app/controllers/translations_controller.rb
```diff
@@ -10,7 +10,9 @@ def show
     if @translation.is_published
       redirect
     else
-      render_not_found
+      @translation.errors.add(:message,
+                              'Translation not found.  Use drafts/ if you\'re looking for an unpublished translation.')
```
soberstadt: Seems like there are a few issues with this. two spaces between the sentences (is that still a debate?), a confusing `/`, and I would just use `"` to avoid escaping.

PR #41, opened 2017-05-12
JSON api improvements + always use loud(!) methods

dsgoers:


app/controllers/application_controller.rb
```diff
@@ -3,7 +3,11 @@
 class ApplicationController < ActionController::Base
   before_action :decode_json_api
 
-  rescue_from ActiveRecord::RecordNotFound, with: :render_not_found
+  rescue_from ActiveRecord::RecordNotFound do |_exception|
+    a = NotFoundError.new
+    a.errors.add(:id, 'Not found.')
```
soberstadt: if you want a move verbose message, you could copy something like: https://github.com/CruGlobal/missionhub-api/blob/master/app/controllers/apis/v4/base_controller.rb#L134-L140

PR #42, opened 2017-05-12
GT 276

dsgoers:


app/controllers/attachments_controller.rb
```diff
@@ -1,6 +1,12 @@
 # frozen_string_literal: true
 
-class AttachmentsController < SecureController
+class AttachmentsController < ApplicationController
+  before_action :authorize!, only: [:create, :update, :destroy]
+
+  def download
+    redirect_to load_attachment.file.url, status: :found
```
soberstadt: I've never used this found status, do you need it here?

app/models/attribute.rb
```diff
@@ -1,29 +1,25 @@
...
 
   def set_defaults
     self.is_translatable ||= false
   end
 
-  def duplicate_keys # see note on same method in attachment.rb
-    return unless resource.attachments.find_by(key: key).present?
-    raise 'Key is current used by an Attachment.'
+  def lower_key
+    self.key = key.downcase if key.present?
```
soberstadt: `self.key = key&.downcase` should also work if you want it.

app/controllers/attachments_controller.rb
dsgoers: No, but the app developers requested it.

soberstadt: Huh, ok

dsgoers: To differentiate between other types of redirect.

PR #43, opened 2017-05-15
Revert to earlier version of AMS because type shouldn't be plural in …

dsgoers:
…a relationship


dsgoers:
cc @frett

soberstadt:
Otherwise you could utilize something like what I have here: https://github.com/CruGlobal/missionhub-api/blob/master/app/serializers/v4/base_serializer.rb#L4-L15

I personally decided to stay on 0.10.6 because I desired the performance boost from not loading the belongs_to object into memory, just so we can get it's id and type.

PR #44, opened 2017-05-15
Add tabs container

pablomarti:


pablomarti:
Currently I'm using the XML for making the tabs on the 4 view of KGP (when displaying the circles).

manifest/content.xml
```diff
@@ -73,4 +73,27 @@
...
 
+  <!--
+    Tabs
+      Required childr elements: options, option, tab-content, tab
+      Each option element requires a single content:text child element
+      The number of tab elements should be the same of the number of option element
+  -->
+  <tabs>
+    <options>
+      <option><content:text i18n-id="{{uuid}}">Option A</content:text></option>
```
frett: can we simplify this XML by integrating the options directly into the tab node itself. something like:

```
<tabs>
  <tab>
    <label>
      <text i18n-id="{{uuid}}">Option A</text>
    </label>
    <image />
    <text />
  </tab>
</tabs>

Comment (frett) [inline manifest/content.xml]: where `<label>` defines the label within the tab control itself?

Comment (pablomarti) [inline manifest/content.xml]: Ok

**Comment by pablomarti** [comment]
Just made the change @frett

File: manifest/content.xml [hunk 2]
```diff
@@ -73,4 +73,29 @@
...
   </form>
 
+  <!--
+    Tabs
+      Required childr elements: options, option, tab-content, tab
+      Each option element requires a single content:text child element
+      The number of tab elements should be the same of the number of option element
+  -->
+  <tabs>
+    <tab-content>
```
frett: `<tab-content>` shouldn't be needed anymore.

pablomarti: You are right! just removed it

manifest/content.xml
```diff
@@ -73,4 +73,27 @@
     <button />
   </form>
 
+  <!--
+    Tabs
+      Required childr elements: options, option, tab-content, tab
```
ryancarlson: typo: "childr"

manifest/content.xml
```diff
@@ -73,4 +73,27 @@
...
+      Each option element requires a single content:text child element
+      The number of tab elements should be the same of the number of option element
+  -->
+  <tabs>
+    <tab>
+      <label>
+        <text i18n-id="{{uuid}}">Option A</text>
```
ryancarlson: Will this be namespaced like the other text elements? Or is it different because it's not "body" text?

frett: it is namespaced, the default XML namespace on this document is the content namespace :)

pablomarti: Not sure about it, will every text change ? @frett

manifest/content.xml
```diff
@@ -73,4 +73,27 @@
     <button />
   </form>
 
+  <!--
+    Tabs
+      Required childr elements: options, option, tab-content, tab
+      Each option element requires a single content:text child element
+      The number of tab elements should be the same of the number of option element
+  -->
+  <tabs>
```
ryancarlson: iOS supports 2 or more options on a segmented control which is what we would use here. Should we suggest a soft maximum number though to ensure usability. 2 or no more than 3?

manifest/content.xml
frett: all of the XML elements in this specific file are in the same xml namespace as `content:*` in other documents. XML namespace prefixes are only used to lookup the namespace URI in the xml document they are in. Prefixes can be any string you want within that xml file as long as you assign them to the correct namespace URI :)

manifest/content.xml
frett: I haven't looked at the android control I would use for this, but I imagine it would be a similar situation for me. I think adding suggestions would be beneficial. I would want it to only be a suggestion, not a requirement though.

ryancarlson: Agreed.., let's say recommended maximum is two.

ryancarlson:
Those are all my questions.

frett:
the XML itself looks good to me, so @pablomarti go ahead and start using this XML as the basis for tabs. I still want a bit more documentation on what `<label>` means and the suggestion of no more than 3 tabs.

pablomarti:
Cool @frett !

frett:
I'm going to update this to represent what actually landed in the XML Schema :)

PR #46, opened 2017-05-15
Add page order

dsgoers:


lib/s3_util.rb
```diff
@@ -41,7 +41,7 @@ def build_zip
   end
 
   def add_pages(zip_file, pages_node)
-    @translation.resource.pages.each do |page|
+    @translation.resource.pages.order(position: :asc).each do |page|
```
knutsenm: Since you allow nulls, you might want to get explicit here about whether this sorts nulls first or last, and/or test for what you expect to happen:
http://stackoverflow.com/questions/5826210/rails-order-with-nulls-last

dsgoers:
Is this good to go?

db/migrate/20170515171542_page_order.rb
```diff
@@ -0,0 +1,12 @@
+class PageOrder < ActiveRecord::Migration[5.0]
+  def change
+    add_column :pages, :position, :integer, null: true
+
+    Page.all.each do |p|
+      p.position = p.id #assign it a value we know is unique
```
knutsenm: Very clever!

dsgoers:
Thanks for the review!

PR #48, opened 2017-05-16
GT 288

dsgoers:
Dynamically generate TranslationElements based on given page structure.


config/initializers/rollbar.rb
```diff
@@ -30,6 +30,7 @@
   #
   # You can also specify a callable, which will be called with the exception instance.
   config.exception_level_filters['ActionController::RoutingError'] = 'ignore'
+  config.exception_level_filters['AbstractController::ActionNotFound'] = 'ignore'
```
knutsenm: Hmm; @soberstadt, is it wise to ignore all these?

app/controllers/pages_controller.rb
```diff
@@ -1,9 +1,17 @@
 # frozen_string_literal: true
 
 class PagesController < SecureController
+  def create
+    p = Page.create!(params.require(:data).require(:attributes)
```
knutsenm: The API docs warn against require( ).require( ) and suggest an alternative

http://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-require

...I don't understand exactly what is being said there, but I don't find any examples of require( ).require( ) in our code either.

app/controllers/pages_controller.rb
```diff
@@ -1,9 +1,17 @@
...
+                       .permit([:filename, :structure, :resource_id, :position]))
```
knutsenm: You don't have to wrap the params to permit( ) in an array; none of the examples do.

You should consider configuring rubocop; it would complain about indentation here and it might alert about the array as well.

app/controllers/pages_controller.rb
```diff
@@ -1,9 +1,17 @@
...
+
+    response.headers['Location'] = "pages/#{p.id}"
+    render json: p, status: :created
+  end
+
   def update
     page = Page.find(params[:id])
-    page.update!(structure: params[:data][:attributes][:structure])
+    page.update!(params.require(:data).require(:attributes).permit(:structure))
```
knutsenm: Same caution as above, about chaining .require

app/models/page.rb
```diff
@@ -9,4 +9,21 @@ class Page < ActiveRecord::Base
...
   validates :resource, presence: true
   validates :position, presence: true, uniqueness: { scope: :resource }
+
+  after_save :upsert_translation_elements
+
+  private
+
+  def upsert_translation_elements
+    Nokogiri::XML(structure).xpath('//content:text[@i18n-id]').each do |node|
+      onesky_phrase_id = node['i18n-id']
```
knutsenm: OK, but is this guaranteed to be present?

dsgoers: No it's not; I realize I don't have any error checking there right now... and I'm fine with that for now

dsgoers: The only reason it wouldn't be present is if some bad XML was sent in.  Ideally it would always be present

app/controllers/pages_controller.rb
dsgoers: I have Rubocop configured in this project, but it doesn't seem to care about the indentation here.  I have my continuation indent set to 2, so that's why it's indented two spaces.

app/controllers/pages_controller.rb
dsgoers: The reason I'm doing this is because with JSON api the request usually looks like this:

app/models/page.rb
dsgoers: I have to add some code to verify the XML is correct in a different JIRA task... so I will address this then 😄

frett: to add to this, the XPath expression that iterates over nodes only selects nodes that have the `i18n-id` attribute.

PR #51, opened 2017-05-17
Xml schema

frett:
This PR adds an XML Schema for tract pages. To validate the sample kgp pages include with this repo, checkout the branch and run this command: `xmllint --noout --schema manifest/xsd/tract.xsd manifest/kgp/kgp-*.xml`


manifest/kgp/kgp-5.xml
```diff
@@ -0,0 +1,115 @@
...
+            </content:input>
+            <content:input type="text" name="email">
+                <content:label>
+                    <content:text i18n-id="{{uuid}}">Email</content:text>
+                </content:label>
+                <content:placeholder>
+                    <content:text i18n-id="{{uuid}}">Email</content:text>
+                </content:placeholder>
+            </content:input>
+            <content:button type="event" value="followup:send,send-information-modal">
```
pablomarti: There is no explicit listener for `followup:send` in this XML, so, this listener should be kind of hardcoded in the app, right? @frett

manifest/xsd/content.xsd
```diff
@@ -0,0 +1,161 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
pablomarti: For what are we going to use this file?

manifest/xsd/tract.xsd
```diff
@@ -0,0 +1,202 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
pablomarti: For what are we going to use this file?

manifest/kgp/kgp-5.xml
frett: there is a namespace concept we incorporated in event id's, this is an external event from the package content and was the same event used in previous apps when submitting a followup form.

manifest/xsd/tract.xsd
frett: These are the actual schemas for validating tract xml. If we generate something that doesn't follow our spec these schemas will be able to report what is invalid about the XML. I want these to be utilized in the API when creating/managing pages so we can guarantee that we will be working against expected XML and not introduce bugs that end up becoming part of the spec.

manifest/kgp/kgp-5.xml
frett: long term goal for events was the possibility of triggering an event in one tool and having it jump to a listener in a different tool. This is not required right now, but we designed for it originally, and utilize it to capture forms being submitted.

manifest/xsd/tract.xsd
pablomarti: Awesome!

pablomarti: Hey what tool do you use for doing that validation?

frett: These won't be present in the mobile app, or need to be used by the mobile app at all, this will be something for the API, and something to check our sample XML against as we are creating it

frett: on a mac I'm running `xmllint`

frett: `xmllint --noout --schema manifest/xsd/tract.xsd manifest/kgp/kgp-*.xml`

manifest/kgp/kgp-5.xml
pablomarti: I like that! Even possible for deep-linking (for future)

PR #52, opened 2017-05-17
Auth tests

dsgoers:


dsgoers:
by public, do you mean the endpoint doesn't require authorization?

soberstadt:
Yes, that is what I meant

dsgoers:
Yes, there's a lot of public endpoints.

dsgoers:
I see what you're saying... maybe later I'll add that.  For now the most important thing is making sure secured endpoints are actually secure.

PR #53, opened 2017-05-23
GT 278

dsgoers:
Add compatibility with EveryStudent content.  This content will be sent to the API translated, rather than use OneSky for translating each phrase like the tracts do.


dsgoers:
@Omicron7 did you see this?

dsgoers:
thanks!

PR #58, opened 2017-05-30
GT 290

dsgoers:
We need to validate the XML in Pages and Custom Pages.  The validation to be used depends on the type of Resource (ie tracts have different layouts than articles)


app/models/resource_type.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class ResourceType < ActiveRecord::Base
+  validates :name, presence: true, uniqueness: true
+  validates :dtd_file, presence: true
+
+  after_validation :dtd_file_exists
```
twinge: Tell me more about this. Why is it an `after_validation`?

app/models/resource_type.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  private
+
+  def dtd_file_exists
+    return if File.exist?("public/xmlns/#{dtd_file}")
+    raise "ResourceType with name: #{name} does not have valid DTD file in 'public/xmlns/'."
```
twinge: Wondering if this message might be a bit misleading in that it says there's no a 'valid' DTD file. Doesn't seem like any validation of the contents of the DTD file is happening here.

app/models/resource_type.rb
dsgoers: Since it is doing validation, I wanted to use one of the `x_validation` callbacks as opposed to `x_save`.  I want it to be done on creates and updates.

dsgoers: Do you think it should be something different?

dsgoers:
thanks!

PR #62, opened 2017-06-02
Allow views to be updated w/o authorization token

ryancarlson:


ryancarlson:
@dsgoers guessing at if this is the right way to fix this issue, but it seems reasonable :)

dsgoers:
run `rubocop -a`

PR #63, opened 2017-06-02
GT 310

dsgoers:
This table is redundant because we generate what we need dynamically


db/migrate/20170530201527_remove_one_sky_phrases_from_database.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveOneSkyPhrasesFromDatabase < ActiveRecord::Migration[5.0]
+  def change
+    drop_table :onesky_phrases do |t|
+      t.string :text, null: false
+      t.string :onesky_id, index: { unique: true }, null: false
+      t.references :page, null: false
```
knutsenm: Does the above make this reversible? Cool.

app/models/onesky_phrase.rb
```diff
@@ -1,20 +1,10 @@
 # frozen_string_literal: true
 
-class OneskyPhrase < ActiveRecord::Base
-  belongs_to :page
+class OneskyPhrase
```
knutsenm: Why is this class needed at all?

db/migrate/20170530201527_remove_one_sky_phrases_from_database.rb
dsgoers: Yup 😄

dsgoers:
thanks!!

PR #64, opened 2017-06-02
Add a helper

dsgoers:


Omicron7:
If I remember, files inside the `helpers` directory are by convention view helpers in rails and may not be auto loaded when you want to use them. I think we ran into issue with that when you started this app.

PR #65, opened 2017-06-02
Update docker build

Omicron7:
This updates the build to use the newer 2.3 base image and the updates included in it. Rails builds now run inside docker instead of on jenkins.


dsgoers:
Why don't we need to cache gems anymore?

Omicron7:
The new build process as implemented by Josh doesn't use them. It was unlikely it was using them before as --path arguments sent to different versions of bundler were different. It also adds 25+ MB to the git repo and doesn't seem to speed up the build that much.

PR #67, opened 2017-06-05
Initial manifest XSD

frett:


lib/s3_util.rb
```diff
@@ -42,7 +42,9 @@ def build_zip
   end
 
   def find_or_create_manifest_node
-    return @document.xpath('/manifest').first if @translation.resource.manifest.present?
+    if @translation.resource.manifest.present?
+      return @document.xpath('/m:manifest', 'm' => 'https://mobile-content-api.cru.org/xmlns/manifest').first
```
frett: +1, I like the explicit specification of the manifest namespace here :)

dsgoers: We have to in order for `xpath` to work

PR #69, opened 2017-06-06
GT 291

dsgoers:
We need to add the translated name of the resource to the manifest file, if the manifest file is present.  If you have any suggestions for refactoring up `s3_util.rb`, I'm open to suggestions.


dsgoers:
thanks!!

PR #70, opened 2017-06-06
Handle invalid records and return as JSON api

dsgoers:


app/controllers/application_controller.rb
```diff
@@ -12,6 +12,10 @@ class ApplicationController < ActionController::Base
     render_error(ApiError.new(:id, exception.message), :bad_request)
   end
 
+  rescue_from ActiveRecord::RecordInvalid do |exception|
```
knutsenm: You can rescue_from a list of exceptions, & thus could combine this with the above.

PR #71, opened 2017-06-06
add navbar-color and navbar-control-color as manifest attributes

frett:


frett:
@pablomarti thoughts on using this to define the navigation bar colors?

pablomarti:
@frett I like it, it is better to make a difference between the navigation color and the color that is going to be used for the components.

PR #72, opened 2017-06-06
GT 335

dsgoers:
Store translatable attributes in OneSky

### DIFF
--- a/lib/page_util.rb
+++ b/lib/page_util.rb
@@ -27,13 +27,6 @@ class PageUtil
 
   private
 
-  def push_name_description
-    filename = 'name_description.xml'
-    phrases = { name: @resource.name, description: @resource.description }
-
-    push_page(phrases, filename, false)
-  end
-
   def push_all_resource_pages(keep_existing_phrases)
     @resource.pages.each do |page|
       phrases = {}
@@ -41,6 +34,13 @@ class PageUtil
 
       push_page(phrases, page.filename, keep_existing_phrases)
     end
+  end
+
+  def push_name_description
+    filename = 'name_description.xml'
+    phrases = { name: @resource.name, description: @resource.description }
+
+    push_page(phrases, filename, false)
   end
 
   def push_page(phrases, filename, keep_existing_phrases)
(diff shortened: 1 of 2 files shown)


lib/page_util.rb
```diff
@@ -43,6 +37,18 @@ def push_all_resource_pages(keep_existing_phrases)
...
   end
 
+  def push_name_description
+    phrases = { name: @resource.name, description: @resource.description }
+    push_page(phrases, 'name_description.xml', false)
+  end
+
+  def push_translatable_attributes
+    phrases = {}
+    @resource.resource_attributes.where(is_translatable: true).each { |a| phrases[a.key] = a.value }
```
twinge: There's a slick ruby syntactical sugar way to do this:
```ruby
phrases = Hash[@resource.resource_attributes.where(is_translatable: true).pluck(:key, :value)
```
The `pluck` method selects just the columns listed, and loads them into an array of arrays (instead of `ResourceAttribute` objects). `Hash[]` converts an array of arrays into a hash using the first value of each pair as the key and the second value as the value.

spec/page_util_spec.rb
```diff
@@ -44,12 +49,23 @@
...
-    page_one = Page.new(filename: filename_1, structure: structure_1)
-    page_two = Page.new(filename: filename_2, structure: structure_2)
+  let(:resource) do
+    attributes = [attr_1, attr_2, Attribute.new(key: 'bill', value: 'test 3', is_translatable: false)]
+
+    pages = [Page.new(filename: filename_1, structure: structure_1, position: 1),
+             Page.new(filename: filename_2, structure: structure_2, position: 2)]
 
-    resource = Resource.new(pages: [page_one, page_two], onesky_project_id: 1)
+    Resource.create!(pages: pages,
```
twinge: This would be a good time to use the `factory_girl` library: https://github.com/thoughtbot/factory_girl

I should probably have brought that up in our earlier conversation about mocks. For the times when a mock won't do and you need a real object, factory_girl provides a nice organized way to build those 'fixtures'. Not saying you have to introduce that at this point in the project (depending on how pressed you are for time), but it's what we use on other projects.

twinge:
I'm impressed with how quickly you've picked up all the rails syntax and specs. Good job!

PR #76, opened 2017-06-07
GT 324/325

dsgoers:
Here's what I have so far, what do you think @ryancarlson ?

### DIFF
--- a/app/models/follow_up.rb
+++ b/app/models/follow_up.rb
@@ -3,8 +3,14 @@
 class FollowUp
   include ActiveModel::Validations
 
-  attr_accessor :email, :language
+  attr_accessor :email, :name, :language_id
 
   validates :email, presence: true
   validates :language, presence: true
+
+  def initialize(email, language_id, name = nil)
+    self.email = email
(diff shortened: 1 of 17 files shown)


app/models/follow_up.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  attr_accessor :email, :name, :language, :destination
+
+  validates :email, presence: true, email_format: { message: 'Invalid email address' }
+  validates :language, presence: true
+  validates :destination, presence: true
+
+  def initialize(email, language_id, destination_id, name = nil)
+    self.email = email
+    self.language = Language.find(language_id)
```
twinge: Just making sure you know that the `find` method will throw an exception if for some reason it can't find a record with that id. From what I can tell, the id is coming in from an API call, which means it could be a bad id. You might want to rescue the possible exception in the controller and respond with a friendly error message. Otherwise they'll just get a 500 and think that the system failed when really it was bad input.

dsgoers: I handle that in `application_controller.rb`

PR #77, opened 2017-06-07
Onesky

dsgoers:
We don't need these going to Rollbar.


dsgoers:
On second thought, I would rather these go to Rollbar since they are not user errors.

PR #81, opened 2017-06-09
Refactor some validation and error handling code

dsgoers:


app/controllers/application_controller.rb
```diff
@@ -3,13 +3,25 @@
 class ApplicationController < ActionController::Base
   before_action :decode_json_api
 
-  rescue_from ActiveRecord::RecordNotFound do |_exception|
-    render_error(ApiError.new(:id, 'Not found.'), :not_found)
+  rescue_from ActiveRecord::RecordNotFound do |exception|
+    render_api_error(exception.to_s, :not_found)
```
twinge: Is there a reason for using `exception.to_s` here and `exception.message` in the other cases below? I'm assuming they do the same thing, in which case consistency is nice.

dsgoers:
Wow, that was a fast review

dsgoers:
thanks!!!

twinge:
yw

PR #85, opened 2017-06-09
GT 286

dsgoers:
Check if object exists in S3 before redirecting.  Also remove unnecessary test constants file


dsgoers:
thanks!!

PR #88, opened 2017-06-14
Track the tool code and language in the published manifest

frett:
I want to include the resource code and language in published manifest xml files. This should be a simple change, but not sure how to write a spec to test it.

expected output manifest xml:
```xml
<manifest
    code="kgp"
    language="en" />
```


frett:
closing this for now

PR #96, opened 2017-07-04
Look for a custom page and use it if found.

ryancarlson:


pablomarti:
@ryancarlson @dsgoers  I have been checking the project (and reading Nokogiri doc too), and, I can see:
- CustomPage is only for override the XML, not just the texts, right? So, in that way, yes, CustomPage should be associated to Language instead to Translation
- Right now with the change of Ryan, I believe it is incomplete, because in the model `Translation` there is a method `onesky_translated_page(page_id, strict) ` that eventually will be called, and in that method it is looking specifically for a `Page`, so, it needs to change to look for a `CustomPage` if needed, is that right? If that is the case, it is needed to have an attribute `filename` in `CustomPage` model, so that may require additional changes for allowing that attribute

PR #97, opened 2017-07-05
Make CustomPages a child of Languages instead of Translations

dsgoers:


dsgoers:
thanks!!

ryancarlson:
@twinge will the migration run automagically?

twinge:
yes

PR #99, opened 2017-07-05
[WIP] GT-360: Support i18n-id replacement for attribute nodes

pablomarti:


pablomarti:
Will close this PR because there are some commits that should not be there

PR #101, opened 2017-07-06
Allow for a set of language IDs to be passed in to create multiple dr…

ryancarlson:
…afts in one request


ryancarlson:
@soberstadt test worked in stage. okay to merge?

app/controllers/drafts_controller.rb
```diff
@@ -11,11 +11,14 @@ def show
 
   def create
     resource = load_resource
-    existing_translation = Translation.latest_translation(resource.id, language_id)
 
-    d = existing_translation.nil? ? resource.create_new_draft(language_id) : create_new_version(existing_translation)
-    response.headers['Location'] = "drafts/#{d.id}"
-    render json: d, status: :created
+    create_draft(resource, language_id) unless language_id.nil?
```
soberstadt: I would probably prefer `if language_id.present?`, but it's up to your preference.

ryancarlson: that makes sense

PR #105, opened 2017-07-07
Support paragraph as an element in a form

ryancarlson:


ryancarlson:
@frett I think this does it?

public/xmlns/content.xsd
```diff
@@ -322,6 +322,7 @@
             <xs:element ref="content:link" />
             <xs:element ref="content:form" />
             <xs:element ref="content:input" />
+            <xs:element ref="content:paragraph" />
```
frett: I would bump it up to the beginning of the list so that it matches the order of the element definitions. it will make it easier to keep organized if we ever add on more content types

PR #107, opened 2017-08-07
GT 381

dsgoers:
EveryStudent article schema

### DIFF
--- a/public/xmlns/article.xsd
+++ b/public/xmlns/article.xsd
@@ -3,6 +3,7 @@
   <xs:element name="article">
     <xs:complexType>
       <xs:sequence>
+        <xs:element name="category" type="xs:string" minOccurs="0" maxOccurs="unbounded"/>
         <xs:element name="title" type="xs:string"/>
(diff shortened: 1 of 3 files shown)


public/xmlns/article.xsd
```diff
@@ -1,4 +1,16 @@
 <?xml version="1.0" encoding="utf-8"?>
-<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
-  <xs:element name="item"/>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           attributeFormDefault="unqualified" elementFormDefault="qualified"
+           targetNamespace="https://mobile-content-api.cru.org/xmlns/article">
+  <xs:element name="article">
+    <xs:complexType>
+      <xs:sequence>
+        <xs:element name="category" type="xs:string" minOccurs="0" maxOccurs="unbounded"/>
```
dsgoers: Each of these elements will need to be translated manually according to the current spec

PR #108, opened 2017-08-08
GT 381

dsgoers:
This PR addresses an issue with EveryStudent content.  With our tracts such as GodTools/Satisfied, they are built in a base language (English) and all the translated versions are based on that.  We could not use the same methods for EveryStudent content because there is no base language; we simply want to store as many articles as we have for each language.  Therefore TranslatedPages will not be related to Pages/Translations as the latter classes are only for content that is generated dynamically.

@frett fyi


dsgoers:
thanks!!

PR #113, opened 2017-08-17
Allow creation of Resources through endpoint

dsgoers:


app/controllers/resources_controller.rb
```diff
@@ -39,6 +44,6 @@ def load_resource
   end
 
   def permitted_params
-    [:name, :abbreviation, :manifest, :onesky_project_id, :system_id, :description]
+    [:name, :abbreviation, :manifest, :onesky_project_id, :system_id, :description, :resource_type_id]
```
knutsenm: Does it matter that you are now allowed to change resource_type_id on update?

dsgoers: Yea I was thinking about that and I don't see any issues right now.

dsgoers:
@knutsenm please review the new commit 😄

config/routes.rb
```diff
@@ -3,7 +3,7 @@
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
 
   resources :systems, :auth, :translations, :drafts, :resources, :custom_pages, :languages, :attributes,
-            :translated_attributes, :pages, :views, :attachments, :translated_pages, :follow_ups
+            :translated_attributes, :pages, :views, :attachments, :translated_pages, :follow_ups, :resource_types
```
knutsenm: Should this one be `, only: [:index]` ?

If so, are there any other routes you should be restricting?

dsgoers:
thanks!!

config/routes.rb
dsgoers: I think it's OK like that.  404 is the response if an unimplemented method is called

PR #115, opened 2017-08-17
Show latest draft or translation

dsgoers:
@frett do you have any problems with this?

### DIFF
--- a/app/models/resource.rb
+++ b/app/models/resource.rb
@@ -35,7 +35,7 @@ class Resource < ActiveRecord::Base
   end
 
   def latest_translations
-    latest('is_published = true and')
+    latest(true)
   end
 
   def latest_drafts_translations
@@ -51,7 +51,7 @@ class Resource < ActiveRecord::Base
   private
 
   # returns the Translation with the highest version for each Language and this Resource
-  def latest(is_published = nil)
+  def latest(is_published = [true, false])
     Translation.joins("inner join (#{latest_versions(is_published).to_sql}) as max_table
                        on translations.version = max_table.max_version
                        and translations.language_id = max_table.language_id
@@ -61,8 +61,7 @@ class Resource < ActiveRecord::Base
   # returns the highest version for each Language and this Resource
   def latest_versions(is_published)
     Translation.select(:language_id, :resource_id, 'max(version) as max_version')
-               .from('translations')
-               .where("#{is_published} resource_id = #{id}")
+               .where(resource_id: id, is_published: is_published)
                .group(:language_id, :resource_id)
   end
 end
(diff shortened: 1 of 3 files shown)


dsgoers:
What I was thinking might be a problem is that we more easily expose Drafts to non-admins, but I don't think it's that big of a deal.  They still need to authenticate to see the translated text

dsgoers:
Furthermore, Drafts were already exposed to non-admins through the `translations` relationship.  If we did want to make sure Drafts were not exposed at all to non-admins I could do that, but it would take a little work.

app/models/resource.rb
```diff
@@ -34,18 +34,28 @@ def create_new_draft(language_id)
...
   end
 
-  # Returns the latest translation for each language.  In more detailed terms, the inner query returns the highest
-  # version for the resource/language combination. The outer query then finds the translation matching that version.
   def latest_translations
-    Translation.find_by_sql("select T.* from translations T inner join (select language_id, resource_id, max(version) as
-                            max_version from translations where is_published = true and resource_id = #{id} group
-                            by language_id, resource_id) as max_table on T.version = max_table.max_version and
-                            T.language_id = max_table.language_id and T.resource_id = max_table.resource_id")
+    latest('is_published = true and')
```
twinge: This feels a little dirty. I think what you're trying to accomplish in the `#latest` method below can be accomplished more cleanly by putting that subquery into the `#from` method: http://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-from

app/models/resource.rb
```diff
@@ -34,18 +34,28 @@ def create_new_draft(language_id)
...
+  end
+
+  def latest_drafts_translations
+    latest(nil)
```
twinge: Rather than pass `nil` here, set `nil` as the default in the method definition below:

```ruby
def latest(is_published = nil)
```

spec/models/resource_spec.rb
```diff
@@ -38,6 +38,19 @@
     end
   end
 
+  context 'returns latest translation (published or not) for each language' do
```
twinge: The best part about having these tests is that they should still pass after you refactor the `#latest` method.

frett:
I don't have any issues with drafts being publicly readable

dsgoers:
@twinge is this good to go?

app/models/resource.rb
```diff
@@ -34,18 +34,34 @@ def create_new_draft(language_id)
...
   def total_views
     views.all.sum(:quantity)
   end
 
   delegate :name, to: :resource_type, prefix: true
+
+  private
+
+  # returns the Translation with the highest version for each Language and this Resource
+  def latest(is_published = [true, false])
```
twinge: Just making sure I understand your intent here. By default you want to select everything regardless of if is_published is true or false? If so, this is fine, just making sure.

dsgoers: That is correct.

dsgoers:
thanks!!

PR #116, opened 2017-08-23
Don't merge parameters if body is nil

dsgoers:
I ran into this with a DELETE request.


dsgoers:
More specifically, the problem was the object returned by `request.body` had no `string` method.  It was `Puma::NullIO`

dsgoers:
thanks!!

PR #117, opened 2017-09-01
Order latest translations & drafts by language name

dsgoers:


dsgoers:
bump 😄

app/models/resource.rb
```diff
@@ -56,6 +56,7 @@ def latest(is_published = [true, false])
                        on translations.version = max_table.max_version
                        and translations.language_id = max_table.language_id
                        and translations.resource_id = max_table.resource_id")
+               .includes(:language).order('languages.name ASC')
```
twinge: 'ASC' is the default, you don't need to specify it.

dsgoers: Yea thanks, I just like to be specific 😄

dsgoers:
thanks!!

PR #118, opened 2017-09-13
Sort languages and resources by name ascending

dsgoers:


spec/acceptance/languages_controller_spec.rb
```diff
@@ -15,6 +15,12 @@
 
       expect(status).to be(200)
     end
+
+    it 'sorts by name ascending', document: false do
+      do_request
+
+      expect(JSON.parse(response_body)['data'][0]['id']).to eq(4.to_s)
```
soberstadt: I think I prefer `'4'` instead of `4.to_s`

dsgoers:
thanks!!

PR #122, opened 2017-09-18
codecov test

dsgoers:


dsgoers:
This was only for testing Codecov commenting has been turned off, which apparently failed

PR #124, opened 2017-09-20
Save follow ups in the database

dsgoers:
I decided this would be a good idea after this: https://rollbar.com/Cru/mobile-content-api/items/73/.  The error appears to be a one-off thing, but since the follow-ups aren't stored we have no way of knowing what the actual data was.

Also, in the event of an error, it would be helpful to resend the data to make sure the user gets subscribed as desired

### DIFF
--- a/app/controllers/follow_ups_controller.rb
+++ b/app/controllers/follow_ups_controller.rb
@@ -10,6 +10,6 @@ class FollowUpsController < ApplicationController
   private
 
   def permitted_params
-    data_attrs.permit([:email, :language_id, :destination_id, :name])
+    permit_params(:email, :language_id, :destination_id, :name)
   end
 end
--- a/app/controllers/follow_ups_controller.rb
+++ b/app/controllers/follow_ups_controller.rb
@@ -2,7 +2,7 @@
 
 class FollowUpsController < ApplicationController
   def create
-    f = FollowUp.create!(data_attrs.permit(permitted_params))
+    f = FollowUp.create!(permitted_params)
     f.send_to_api
     head :no_content
   end
@@ -10,6 +10,6 @@ class FollowUpsController < ApplicationController
   private
 
   def permitted_params
-    [:email, :language_id, :destination_id, :name]
+    data_attrs.permit([:email, :language_id, :destination_id, :name])
   end
 end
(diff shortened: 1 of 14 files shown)


app/controllers/follow_ups_controller.rb
```diff
@@ -2,8 +2,14 @@
 
 class FollowUpsController < ApplicationController
   def create
-    f = FollowUp.new(data_attrs[:email], data_attrs[:language_id], data_attrs[:destination_id], data_attrs[:name])
+    f = FollowUp.create!(data_attrs.permit(permitted_params))
```
twinge: Move the `data_attrs.permit()` call down to the permitted params method so that this line becomes 

```ruby
f = FollowUp.create!(permitted_params)
```

app/models/follow_up.rb
```diff
@@ -2,34 +2,24 @@
...
-  def initialize(email, language_id, destination_id, name = nil)
-    self.email = email
-    self.language = Language.find(language_id)
-    self.destination = Destination.find(destination_id)
-    self.name = name
-  end
-
   def send_to_api
-    validate_fields
+    save!
```
twinge: This will result in `save!` being called immediately after a `create!`. I can't remember how smart rails is about this these days. If this results in an extra sql update call, you'll want to add a check here like:
```ruby
save! if changed?
```

app/models/follow_up.rb
```diff
@@ -2,34 +2,24 @@
...
+
+    Rails.logger.info "Sending follow up with id: #{id}."
```
twinge: How often will this get called? Just want to be careful not to log to `info` with anything that's too verbose.

dsgoers: It will be logged every time a follow up is sent

twinge: I get that, I just don't have any sense of how often that will be. Is this something you plan to search for in the logs later? What's the value in logging this? Presumably it's getting logged to the database anyway?

dsgoers: I want to log the ID so I can look it up later if there's a failure

dsgoers:
thanks!!!

PR #125, opened 2017-09-21
Rubocop/refactoring

dsgoers:
1. Address everything in rubocop_todo
2. Refactor some code by removing business rules from the controller


app/models/resource.rb
```diff
@@ -25,13 +25,9 @@ def uses_onesky?
...
-  def create_new_draft(language_id)
-    language = Language.find(language_id)
-
-    # TODO: disable this to prevent the API from overwriting existing translations within OneSky.
-    # TODO: This will probably need to be revisited -DF
-    # PageClient.new(self, language.code).push_new_onesky_translation
-    Translation.create!(resource: self, language: language)
+  def create_draft(language_id)
+    translation = Translation.latest_translation(id, language_id)
+    translation.nil? ? create_first_draft(language_id) : translation.create_new_version
```
twinge: is the call to `.nil?` necessary?

soberstadt: what about `translation&.create_new_version || create_first_draft(language_id)`?

twinge: That feels dirty to me

soberstadt: You're right. `translation ? translation.create_new_version : create_first_draft(language_id)` then

dsgoers:
Fixed.

dsgoers:
@soberstadt

app/controllers/drafts_controller.rb
```diff
@@ -11,10 +11,9 @@ def show
 
   def create
     resource = load_resource
+    language_ids = [].push(data_attrs[:language_ids]).push(data_attrs[:language_id])
```
soberstadt: This line isn't my favorite line ever. It's not exactly clear what you're expecting here.

soberstadt: I would have approved this PR if you hadn't added this commit.

dsgoers: So we can handle two types of input: `language_id: 1` or `language_ids: [1, 2, 3]`

dsgoers: I will change it to only support the latter

dsgoers:
I moved the last commit to a different branch.  I will merge it later since the fix will require front end changes

PR #127, opened 2017-10-10
Simultaneous publishing of drafts

dsgoers:


dsgoers:
This isn't a high priority, but would be helpful.

dsgoers:
thanks!!

PR #129, opened 2017-11-15
add support for defining text-style="bold"

frett:
`text-style` is a token list, so multiple space-separate styles could be defined. currently we only define the `bold` style.

example text node:
```xml
<content:text text-style="bold">Bold Text</content:text>
<content:text text-style="italic underline">Underlined Italic text</content:text>
```

### DIFF
--- a/public/xmlns/content.xsd
+++ b/public/xmlns/content.xsd
@@ -96,6 +96,8 @@
     <xs:simpleType name="textStyle">
         <xs:restriction base="xs:token">
             <xs:enumeration value="bold" />
+            <xs:enumeration value="italic" />
+            <xs:enumeration value="underline" />
         </xs:restriction>
     </xs:simpleType>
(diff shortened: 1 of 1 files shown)


public/xmlns/content.xsd
```diff
@@ -93,6 +93,15 @@
         </xs:restriction>
     </xs:simpleType>
 
+    <xs:simpleType name="textStyle">
+        <xs:restriction base="xs:token">
+            <xs:enumeration value="bold" />
```
ryancarlson: are there any besides **bold** that it makes sense to add now while we're at it?

frett: `italic`? or should it be `italics`

frett: `italic`

ryancarlson: _italic_ and underline ?

PR #130, opened 2017-11-16
Rubocop

dsgoers:
Address https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-8418


.rubocop_todo.yml
```diff
@@ -1,7 +1,101 @@
...
+# Offense count: 1
+# Configuration parameters: ExpectMatchingDefinition, Regex, IgnoreExecutableScripts, AllowedAcronyms.
+# AllowedAcronyms: CLI, DSL, ACL, API, ASCII, CPU, CSS, DNS, EOF, GUID, HTML, HTTP, HTTPS, ID, IP, JSON, LHS, QPS, RAM, RHS, RPC, SLA, SMTP, SQL, SSH, TCP, TLS, TTL, UDP, UI, UID, UUID, URI, URL, UTF8, VM, XML, XMPP, XSRF, XSS
+Naming/FileName:
+  Exclude:
+    - 'Guardfile'
+
+# Offense count: 1
+Performance/UnfreezeString:
+  Exclude:
```
twinge: Why are you setting up all of these excludes (here and below) instead of fixing the files? I'm unclear as to what the point of upgrading rubocop is if the only file you modified as a result of the new rules was Guardfile.

dsgoers: I upgraded Rubocop because of a security vulnerability.  I don't have the time right now to fix all the new offenses it's complaining about or turn off the specific cops.

twinge:
This is the only CVE I could find for rubocop: http://www.cvedetails.com/cve/CVE-2017-8418/

Was it that or something else? That CVE doesn't justify the upgrade (IMO). It doesn't present any real security threat with how we do things.

The "I don't have time to fix them now" line often turns into things never getting addressed. Better to wait to do the upgrade until you have time to address the new issues.

dsgoers:
OK, we can ignore this for now.  However, while I understand the concern with putting issues off, I would still say it should be OK to put Rubocop offenses on the back burner as they aren't "real" issues.

PR #131, opened 2017-12-22
Update tract.xml

ryancarlson:
add documentation for page.card-text-color attribute


frett:
the main documentation in the XSD: https://github.com/CruGlobal/mobile-content-api/blob/master/public/xmlns/tract.xsd#L66

PR #132, opened 2017-12-22
Update manifest.xml

ryancarlson:
Add documentation for navbar properties options


frett:
the main location for the documentation of the spec is in the XSD currently:
https://github.com/CruGlobal/mobile-content-api/blob/master/public/xmlns/manifest.xsd#L46

PR #134, opened 2018-01-23
Upgrade paperclip and rubocop gems

dsgoers:
I know you'll want all those rubocop offenses fixed, but I don't have time now and I would rather add them to the rubocop TODO then ignore the rubocop vulnerability altogether.


dsgoers:
Tested on stage and it works.

PR #136, opened 2018-02-12
Add button-color attribute to callToAction element

ryancarlson:
Is optional, and must follow rbga(#,#,#,#) format. Colors only the button, not any descendent text elements

### DIFF
--- a/public/xmlns/tract.xsd
+++ b/public/xmlns/tract.xsd
@@ -219,6 +219,11 @@
                         </xs:documentation>
                     </xs:annotation>
                 </xs:attribute>
+                <xs:attribute name="button-color" type="content:colorValue" use="optional">
+                    <xs:annotation>
+                        <xs:documentation>This attribute defines the rgba color of the call to action button.</xs:documentation>
+                    </xs:annotation>
+                </xs:attribute>
             </xs:extension>
         </xs:complexContent>
     </xs:complexType>
--- a/public/xmlns/tract.xsd
+++ b/public/xmlns/tract.xsd
@@ -219,7 +219,8 @@
                         </xs:documentation>
                     </xs:annotation>
                 </xs:attribute>
-                <!-- control-color: DEFAULT "GodTools blue" rgba(59,164,219,1) . -->
+                <!-- control-color: iOS DEFAULT is "GodTools blue" rgba(59,164,219,1) , Android DEFAULT is the primary-color attribute set at the
+                nearest ancestor either page, or manifest. -->
                 <xs:attribute name="control-color" type="content:colorValue" use="optional">
                     <xs:annotation>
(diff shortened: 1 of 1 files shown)


ryancarlson:
@frett @dsgoers thoughts?

public/xmlns/tract.xsd
```diff
@@ -219,6 +219,11 @@
                         </xs:documentation>
                     </xs:annotation>
                 </xs:attribute>
+                <xs:attribute name="button-color" type="content:colorValue" use="optional">
+                    <xs:annotation>
+                        <xs:documentation>This attribute defines the rgba color of the call to action button.</xs:documentation>
```
frett: you should document what the default color would be

frett: based on the android code I default to the page primary-color

public/xmlns/tract.xsd
```diff
@@ -219,6 +219,11 @@
                         </xs:documentation>
                     </xs:annotation>
                 </xs:attribute>
+                <xs:attribute name="button-color" type="content:colorValue" use="optional">
```
frett: how about `control-color` we already use `control` elsewhere for similar functionality. see: https://github.com/CruGlobal/mobile-content-api/blob/master/public/xmlns/manifest.xsd#L53

ryancarlson:
updated, good to go?

public/xmlns/tract.xsd
```diff
@@ -219,7 +219,8 @@
                         </xs:documentation>
                     </xs:annotation>
                 </xs:attribute>
-                <!-- control-color: DEFAULT "GodTools blue" rgba(59,164,219,1) . -->
+                <!-- control-color: iOS DEFAULT is "GodTools blue" rgba(59,164,219,1) , Android DEFAULT is the primary-color attribute set at the
+                nearest ancestor either page, or manifest. -->
```
frett: We will need to standardize this for the spec. I would argue it should default to a theme color defined at a higher level. Also, the default color comment should be included within the `xs: documentation` node

ryancarlson: i'm looking at standardizing it to primary-color now.

frett: good deal :)

ryancarlson: Fixed comment and updated iOS renderer here: 
https://github.com/CruGlobal/godtools-swift/commit/e4f76d49511783d64e9d4d5ebb00393e2cf06fbe

ryancarlson:
PR was approved.. travis failed for unclear reasons and codecov never reported. However codecov is unchanged by this PR

frett:
travis failed because one of the gems in the rails project has a potential vulnerability and needs to be upgraded. not related to your PR :)

dsgoers:
@ryancarlson Codecov didn't report because Travis failed.

PR #138, opened 2018-02-13
Gt 449 right to left attribute

ryancarlson:
Adds "direction" attribute to language. This attribute will help clients know which way text ought to be rendered for a given languages. Defaults to "ltr" unless otherwise specified.

endpoint validates values to be either: "ltr" and "rtl"

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -39,7 +39,7 @@ gem 'aws-sdk'
 gem 'aws-sdk-rails'
 gem 'codecov', require: false
 gem 'ddtrace'
-gem 'nokogiri', '>= 1.8.1'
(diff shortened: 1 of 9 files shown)


ryancarlson:
cc @frett

ryancarlson:
@sheldond Here's a PR that introduces a "direction" attribute on languages served by the mobile-content-api. i mentioned we might do this in our call for knowgod.com. Another use-case came up yesterday for the iOS client, so I went ahead with introducing it.

sheldond:
@ryancarlson Perfect, thanks for the update!

ryancarlson:
thanks for reviewing @knutsenm ! thanks for your input @dsgoers and others

PR #143, opened 2018-04-10
Add tagline

ryancarlson:
A new field from name_description.xml for display on app home screen with each tool.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -27,6 +27,10 @@ Documentation:
 Metrics/LineLength:
   Max: 120
 
+Metrics/ClassLength:
+  Exclude:
+    - app/models/translation.rb
+
 Metrics/BlockLength:
   Exclude:
     - spec/**/*
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -27,10 +27,6 @@ Documentation:
 Metrics/LineLength:
   Max: 120
 
-Metrics/ClassLength:
-  Exclude:
-    - app/models/translation.rb
-
 Metrics/BlockLength:
   Exclude:
     - spec/**/*
--- a/app/models/translation.rb
+++ b/app/models/translation.rb
@@ -102,8 +102,7 @@ class Translation < ActiveRecord::Base
 
     name_desc_onesky if resource.uses_onesky?
 
-    package = Package.new(self)
-    package.push_to_s3
+    Package.new(self).push_to_s3
   end
 
   def name_desc_onesky
--- a/app/models/translation.rb
+++ b/app/models/translation.rb
@@ -112,6 +112,7 @@ class Translation < ActiveRecord::Base
     p = download_translated_phrases('name_description.xml')
     self.translated_name = p['name']
     self.translated_description = p['description']
+    self.translated_tagline = p['tagline']
   end
 
   def download_translated_phrases(page_filename)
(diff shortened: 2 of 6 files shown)


frett:
I know we added support for arbitrary attributes to specific components. I can't remember if they were on Translation objects. If they were, we might be able to use that without needing an additional DB field to track it. Let me see if I can find where we defined them.

frett:
nm, arbitrary attributes are at the resources level based on the API documentation

ryancarlson:
@dsgoers can you review please?

db/schema.rb
```diff
@@ -139,6 +139,7 @@
     t.string  "translated_name"
     t.string  "translated_description"
     t.string  "manifest_name"
+    t.string  "translated_tagline",     default: ""
```
dsgoers: Why did you add a default of empty string?

.rubocop.yml
```diff
@@ -27,6 +27,10 @@ Documentation:
 Metrics/LineLength:
   Max: 120
 
+Metrics/ClassLength:
```
dsgoers: I'd rather the class be refactored than silence the cop.

app/models/translation.rb
```diff
@@ -112,6 +112,7 @@ def name_desc_onesky
     p = download_translated_phrases('name_description.xml')
     self.translated_name = p['name']
     self.translated_description = p['description']
+    self.translated_tagline = p['tagline']
```
dsgoers: Can you add a test for this?

.rubocop.yml
ryancarlson: It's over by one line, and I'm not experienced enough in Ruby/Rails yet to know how to refactor this well. My primary goal is to keep Greg moving

db/schema.rb
ryancarlson: so that { translated_tagline: null } wouldn't come out in the JSON

app/models/translation.rb
ryancarlson: There aren't any existing specs for this class that I could see...

db/schema.rb
dsgoers: Well, it doesn't look like `translated_name` and `translated_description` have a default value...

app/models/translation.rb
dsgoers: https://github.com/CruGlobal/mobile-content-api/blob/master/spec/models/translation_spec.rb#L157

.rubocop.yml
dsgoers: That makes sense, but silencing the check allows for the class to grow indefinitely.

app/models/translation.rb
ryancarlson: ....
````
Offenses:

spec/models/translation_spec.rb:165:7: C: Example has too many expectations [3/2].
      it 'updates from OneSky' do
      ^^^^^^^^^^^^^^^^^^^^^^^^

130 files inspected, 1 offense detected

.rubocop.yml
ryancarlson: https://jira.cru.org/browse/GT-476

db/schema.rb
ryancarlson: Does it hurt having a default value?

dsgoers: I don't think so.

.rubocop.yml
dsgoers: See my commit.

ryancarlson:
good?

PR #145, opened 2018-04-17
Tagline drop default

ryancarlson:


ryancarlson:
@wiseguy16 and I chatted about this, and it makes the client code cleaner to not have the default empty string.

db/migrate/20180416210608_translations_tagline_drop_default.rb
```diff
@@ -0,0 +1,6 @@
+class TranslationsTaglineDropDefault < ActiveRecord::Migration[5.0]
+  def change
+    change_column_default(:translations, :translated_tagline, nil)
+    Translation.where( translated_tagline: '' ).update_all( translated_tagline: nil )
```
dsgoers: `update_all` skips callbacks

ryancarlson: I'm not sure what this means.. can you explain?

ryancarlson:
Thanks boss! can i get a ship it squirrel?

dsgoers:
image

PR #146, opened 2018-04-17
Go back to update_all to avoid callbacks

ryancarlson:
updating all records with callbacks means that for every published translation, we call out to onesky and push to s3. this:
 a) takes way too long
 b) overwrites old version of pushed zips to s3. probably not a huge deal, but the versions are nice


soberstadt:
+1

ryancarlson:
good boss, @dsgoers ?

ryancarlson:
thanks!

PR #148, opened 2018-04-26
attribute keys within a single analytics event should be unique

frett:
This change will prevent analytics attributes from accidentally being repeated within a single event


frett:
@wiseguy16 I'm adding a constraint to make sure we won't have duplicate attribute keys within a single event

PR #149, opened 2018-04-26
require at least 1 system for analytics events

frett:


public/xmlns/analytics.xsd
```diff
@@ -51,10 +51,17 @@
...
+            <xs:attribute name="system" use="required">
                 <xs:annotation>
-                    <xs:documentation>Space separated list of analytics systems this event is for.</xs:documentation>
+                    <xs:documentation>
+                        Space separated list of analytics systems this event is for. At least 1 system is required.
+                    </xs:documentation>
                 </xs:annotation>
+                <xs:simpleType>
+                    <xs:restriction base="analytics:systemType">
+                        <xs:minLength value="1" />
```
frett: this is minimum string length, which ensures the value isn't empty

PR #151, opened 2018-05-30
Debug follow up issue results

ryancarlson:
Pass access secret and id in URL as form data instead of as headers. Growthspaces was updated to handle them in this way.
Log response code for all requests


ryancarlson:
@soberstadt DG is off fishing, can you review this?

ryancarlson:
Thanks!

PR #152, opened 2018-06-06
Update sinatra to fix security vulnerability

dsgoers:


dsgoers:
Looked good on staging.

dsgoers:
thanks!

PR #153, opened 2018-07-02
sort pages for a resource

frett:


spec/page_client_spec.rb
```diff
@@ -53,14 +53,15 @@
...
-                     resource_type_id: 1,
-                     system_id: 1)
+    resource = Resource.create!(abbreviation: 'test',
+                                onesky_project_id: 1,
+                                name: name,
+                                description: description,
+                                resource_attributes: attributes,
+                                resource_type_id: 1,
+                                system_id: 1)
+    resource.pages = pages
```
dsgoers: Why did you need to do this?

frett: it was some sort of weird initialization cycle bug. when the resource xml validation rule ran the resource property on the page was nil.

By assigning the pages afterwards the resource property is correctly set

frett: I'm guessing having the closure defined for pages causes it to be handled slightly differently within AR.

frett:
I'm currently testing on stage

frett:
it's ordering correctly and passing the spec on stage, I'm going to merge :)

PR #156, opened 2018-07-10
Allow images to be restricted to web or mobile devices only

frett:
This will allow us to display different images for web vs mobile. The main impetus for this is that we want to render a different inline image for the knowgod.com web UI.

In order for us to utilize this for the web we will need to update the mobile apps to ignore images that are restricted to web only. Because of this need for an update to our mobile apps we may need to dialog around this and figure out if there might be a better solution available.

Example:
```xml
<content:image resource="both1.png" />
<content:image resource="both2.png" restrictTo="mobile web" />
<content:image resource="mobileInline.png" restrictTo="mobile" />
<content:image resource="webInline.png" restrictTo="web" />
```

### DIFF
--- a/public/xmlns/content.xsd
+++ b/public/xmlns/content.xsd
@@ -130,6 +130,38 @@
         <xs:list itemType="content:listener" />
     </xs:simpleType>
 
+    <xs:simpleType name="deviceType">
+        <xs:restriction base="xs:token">
+            <xs:enumeration value="mobile">
+                <xs:annotation>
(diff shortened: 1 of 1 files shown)


frett:
I am open to other suggestions for how to accomplish this or if another attribute name makes more sense.

frett:
I added sample usage of the attribute to Renderer Tests on the stage environment.

frett:
merging after discussion with Ryan

PR #157, opened 2018-07-26
utilize Hash.to_query to correctly escape query params

frett:
I want to test this on stage

### DIFF
--- a/app/models/follow_up.rb
+++ b/app/models/follow_up.rb
@@ -19,19 +19,22 @@ class FollowUp < ActiveRecord::Base
   private
 
   def body
-    "subscriber[route_id]=#{destination.route_id}&subscriber[language_code]=#{language.code}"\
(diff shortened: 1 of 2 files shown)


app/models/follow_up.rb
```diff
@@ -19,19 +19,22 @@ def send_to_api
...
-  def names
-    return nil if name.nil?
+  def auth_params
+    { access_id: destination.access_key_id, access_secret: destination.access_key_secret }
+  end
 
-    names = name.split(' ')
-    "&subscriber[first_name]=#{names[0]}&subscriber[last_name]=#{names[1]}"
+  def subscriber_params
+    { route_id: destination.route_id, language_code: language.code, email: email }.merge(name_params)
```
frett: do I need to do anything special to deal with the fact that `name_params` might return `nil`?

dsgoers: I'm not sure.  I'd say check the documentation on `merge.`

PR #158, opened 2018-07-31
when creating an empty manifest, utilize the correct namespace

frett:
I'm starting to refactor parts of the manifest/page xml generation process to better support some new manifest features in process. This is just a small correctness fix

### DIFF
--- a/lib/package.rb
+++ b/lib/package.rb
@@ -54,7 +54,7 @@ class Package
   def load_or_create_manifest_node
     return load_manifest if @translation.resource.manifest.present?
 
-    @document.root = @document.create_element('manifest', 'xmlns' => 'https://mobile-content-api.cru.org/xmlns/manifest')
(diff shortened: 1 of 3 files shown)


lib/package.rb
```diff
@@ -54,9 +54,7 @@ def build_zip
   def load_or_create_manifest_node
     return load_manifest if @translation.resource.manifest.present?
 
-    manifest = Nokogiri::XML::Node.new('manifest', @document)
-    @document.root = manifest
-    manifest
+    @document.root = @document.create_element('manifest', 'xmlns' => 'https://mobile-content-api.cru.org/xmlns/manifest')
```
frett: is it possible to define `https://mobile-content-api.cru.org/xmlns/manifest` as a string constant somewhere? where should global constants go and what should be the syntax?

dsgoers: That's a good question.  If we can't find the Rails norm you could just create a constants file within `lib/`

frett: I added it as a constant within XmlUtils for now

PR #159, opened 2018-08-01
Add tool, locale, and type meta-data attributes

frett:
I also refactored out manifest xml modifications to prepare for adding translated categories to the manifest in support of article content


app/models/xml/manifest.rb
```diff
@@ -0,0 +1,63 @@
...
+      title_node = XmlUtil.xpath_namespace(manifest_node, 'manifest:title').first
+      return if title_node.nil?
+
+      name_node = title_node.xpath('content:text[@i18n-id]').first
+      name_node.content = @translation.translated_name
+    end
+
+    def pages_node
+      return @pages if @pages
+      @pages = XmlUtil.get_or_create_child(@document.root, XmlUtil::XMLNS_MANIFEST, 'pages')
```
dsgoers: Won't this always `create?`  Why do you need the `get` part?

app/models/xml/manifest.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+module XML
```
dsgoers: It would be better to have separate tests for this since it's a separate file.

app/models/xml/manifest.rb
frett: we actually ran into an issue where multiple pages nodes existed in the manifest since they were in the raw manifest xml.

app/models/xml/manifest.rb
frett: k, I'll see about splitting out these tests from the package_spec

app/models/xml/manifest.rb
dsgoers: Hmm... can you elaborate a little?

frett: the manifest for the four was entered as:
```
<manifest>
  <title />
  <pages />
</manifest>
```
when the mobile-content-api processed this manifest it created a second `<pages>` node. This broke the web renderer and also isn't valid with the spec.

In addition one of the changes coming for articles is including an `<article:aem-import src="">` node within the `<pages>` node, we haven't fully figured out how we will manage that yet, but initially this will probably be defined in the manifest xml stored in the database.

dsgoers: Ohhh I get it.  Cool!

frett:
I split out some of the manifest specific spec tests.

PR #163, opened 2018-09-10
[GT-509] allow images to trigger content events and for tabs to listen for them

frett:
The theory behind this is to allow a user to click an image and have that trigger an active tab


frett:
@gyasistory bump

frett:
This PR I'm looking for any feedback on the updated XML design. The xml schema is designed in an open and flexible manner so parts can be reused in multiple contexts.

PR #170, opened 2018-12-05
Upgrade Rails to fix CVE

dsgoers:


dsgoers:
It's on stage.  Can you test for a day or two?

frett:
yeah, that'd be good with this upgrade. I know Ann is actually working on stage, so that should be good enough testing for us.

dsgoers:
Great.

PR #174, opened 2019-05-14
[#GT-516] referenced resources

Omicron7:
This PR changes the packaging to only include resources/attachments in the zip that are referenced in the manifest/pages. An Error is raised if a referenced attachment does not exist.

The PR also moves the `lib` folder inside of `app`. This allows autoloading, preloading for production and reloading for development.


Omicron7:
Associated JIRA https://jira.cru.org/browse/GT-516

dsgoers:
Hey Brian, thanks for your work on this!  Unfortunately I'm on full-time MPD right now 😢.  I would suggest assigning this to Josh and/or Frett.

app/lib/package.rb
```diff
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
 
 require 'zip'
-require 'page_client'
-require 'xml_util'
 
 class Package
+  XPATH_RESOURCES = %w[//@background-image
+                       //manifest:manifest/@banner
```
frett: this one should be `//manifest:category/@banner`
https://github.com/CruGlobal/mobile-content-api/blob/master/public/xmlns/manifest.xsd#L140

app/lib/package.rb
```diff
@@ -56,20 +64,25 @@ def add_pages(zip_file, manifest)
...
 
     sha_filename
   end
 
-  def add_attachments(zip_file, manifest)
-    @translation.resource.attachments.where(is_zipped: true).each do |a|
-      Rails.logger.info("Adding attachment with id: #{a.id} to package for translation with id: #{@translation.id}")
+  def add_attachments(zip_file, manifest) # rubocop:disable Metrics/AbcSize
+    @resources.uniq.each do |filename|
+      attachment = @translation.resource.attachments.find_by(file_file_name: filename)
```
knutsenm: Can there be multiple attachments with the same `file_file_name`? The prior `.where` would find them but `.find_by` will not.

frett: There shouldn't be, and if there are the apps wouldn't know what to do with multiple files with the same name anyways

app/lib/package.rb
```diff
@@ -64,7 +75,7 @@ def write_page_to_file(page)
   end
 
   def add_attachments(zip_file, manifest)
-    @translation.resource.attachments.where(is_zipped: true).each do |a|
+    @translation.resource.attachments.where(file_file_name: @resources.uniq).each do |a|
```
knutsenm: Do you truly no longer care about `is_zipped`? Should that attribute go away?

app/lib/package.rb
```diff
@@ -56,20 +64,25 @@ def add_pages(zip_file, manifest)
...
+    determine_resources(document)
     sha_filename = XmlUtil.xml_filename_sha(translated_page)
 
     File.write("#{@directory}/#{sha_filename}", translated_page)
 
     sha_filename
   end
 
-  def add_attachments(zip_file, manifest)
-    @translation.resource.attachments.where(is_zipped: true).each do |a|
```
knutsenm: Do you truly no longer care about is_zipped? Should that attribute go away?

app/lib/package.rb
frett: This logic is meant to completely replaced the functionality `is_zipped` provided. but leaving it in place for now would allow us to rollback if necessary.

config/environments/production.rb
```diff
@@ -41,7 +41,7 @@
   # config.action_cable.allowed_request_origins = [ 'http://example.com', /http:\/\/example.*/ ]
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  # config.force_ssl = true
+  config.force_ssl = true
```
frett: @Omicron7 could this be causing the startup issues on ECS for production?

PR #175, opened 2019-05-17
exclude the lb monitor from the TLS redirect settings

frett:


frett:
This is not currently working on staging.

`config.force_ssl = true` enables TLS redirects, but the load balancer make a non-SSL request to `/monitors/lb` which triggers a redirect instead of a 200 OK response.

We should exclude `/monitors/lb` from the TLS redirect logic, but this PR currently doesn't appear to be working.

PR #177, opened 2019-07-15
GT-517:  Allow manifests to be processed for OneSky translations

kares:
https://jira.cru.org/browse/GT-517

there's some open questions that I do have (will try to point them with code comments).
when some of those assumptions change or more cleanup is preferred this will need more work.
also did a bit of refactoring as I was absorbing code into my brain (happy to revert as/if needed).

### DIFF
--- a/app/lib/one_sky.rb
+++ b/app/lib/one_sky.rb
@@ -1,0 +1,36 @@
+module OneSky
+
+  # @return [Hash] of translated phrases from OneSky
+  def self.download_translated_phrases(filename, project_id:, language_code:)
+    logger.info "Downloading translated phrases for: #{filename} with language: #{language_code}"
+
+    response = RestClient.get "https://platform.api.onesky.io/1/projects/#{project_id}/translations",
+                              params: headers(filename, language_code)
+
+    # NOTE: maybe Error::TextNotFoundError should have been a error under a OneSky namespace
+    raise Error::TextNotFoundError, 'No translated phrases found for this language.' if response.code == 204
+    JSON.parse(response.body)
+  end
+
+  private
+
+  def download_translated_phrases(*args)
+    OneSky.download_translated_phrases(*args)
+  end
+
+  def self.headers(filename, language_code)
+    { api_key: ENV['ONESKY_API_KEY'], timestamp: AuthUtil.epoch_time_seconds, dev_hash: HashUtil.dev_hash,
+      locale: language_code, source_file_name: filename, export_file_name: filename }
+  end
+
+  class << self
+
+    private
+
+    def logger
+      Rails.logger
+    end
+
+  end
+
+end
--- a/app/lib/one_sky.rb
+++ b/app/lib/one_sky.rb
@@ -1,36 +1,62 @@
-module OneSky
+# frozen_string_literal: true
 
+require 'json'
+require 'rest-client'
+require 'auth_util'
+require 'hash_util'
+
+module OneSky
   # @return [Hash] of translated phrases from OneSky
   def self.download_translated_phrases(filename, project_id:, language_code:)
     logger.info "Downloading translated phrases for: #{filename} with language: #{language_code}"
 
     response = RestClient.get "https://platform.api.onesky.io/1/projects/#{project_id}/translations",
-                              params: headers(filename, language_code)
+                              params: headers(language_code).merge(
+                                source_file_name: filename, export_file_name: filename
+                              )
 
     # NOTE: maybe Error::TextNotFoundError should have been a error under a OneSky namespace
     raise Error::TextNotFoundError, 'No translated phrases found for this language.' if response.code == 204
     JSON.parse(response.body)
   end
 
-  private
-
   def download_translated_phrases(*args)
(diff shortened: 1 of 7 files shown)


app/models/xml/manifest.rb
```diff
@@ -12,7 +14,11 @@ def initialize(translation)
 
       manifest_node = @document.root
       add_manifest_metadata(manifest_node)
-      insert_translated_name(manifest_node)
+
+      phrases = manifest_translated_phrases(manifest_node)
+      # TODO: go strict, what about keeping default title (if not translated)?
```
kares: existing specs assumed non-strict translation as the title i18n to exist wasn't 'strongly' required.

app/models/xml/manifest.rb
```diff
@@ -35,12 +43,23 @@ def add_manifest_metadata(manifest_node)
       manifest_node['type'] = @translation.resource.resource_type.name
     end
 
-    def insert_translated_name(manifest_node)
+    def manifest_translated_phrases(manifest_node)
+      phrases = @translation.manifest_translated_phrases
+      # backwards compatibility with manifest name translation:
+      if (i18n_id = title_i18n_id(manifest_node))
+        phrases = phrases.dup
+        phrases[i18n_id] = @translation.translated_name
```
kares: `translation.translated_name` could be dropped but I was unsure so I rather kept it, for now

app/models/xml/manifest.rb
frett: not sure if this is exactly what you are referring to, but if a `<content:text>` node doesn't contain the `i18n-id` attribute it should be left as is. I'm guessing most of our existing manifests don't have the `i18n-id` attribute for the title, which I will make sure we update to include it everywhere.

app/models/xml/manifest.rb
frett: does `translation.translated_name` get stored in the database? If it is, I believe it is returned by API calls and is used by the mobile apps.

app/models/xml/manifest.rb
kares: the assumption here, with existing specs was that given the following : 
```
<?xml version="1.0"?>
<manifest xmlns="https://mobile-content-api.cru.org/xmlns/manifest" xmlns:content="https://mobile-content-api.cru.org/xmlns/content" tool="kgp" locale="en" type="tract">
  <title>
    <content:text i18n-id="89a09d72-114f-4d89-a72c-ca204c796fd9">Knowing God Personally</content:text>
  </title>
</manifest>
```
... the title translation does not need to exist and the package would have been build fine.
if we want to have this 'strict' we simply need to update some package related specs not to assume so

frett: ah, I would like to make it strict, if we need to tweak some specs to make it work, then we should

kares: believe, it's the a pathological case where the `name_description.xml` would not have a **name** present

app/models/xml/manifest.rb
kares: yep, its updated from OneSky (`name_description.xml`'s **name** value) on every translation update

app/models/xml/manifest.rb
frett: ah, then this could be dropped. If it's not in OneSky we should let it raise a missing translation error

frett: oh wait, I though your last comment was on the other thread

kares:
code coverage mostly decreased due refactoring ... so I believe it isn't relevant.

*app/lib/one_sky.rb* and *app/models/xml/translatable.rb* pre pretty much just extractions from existing pieces.

app/lib/one_sky.rb
```diff
@@ -0,0 +1,62 @@
...
+  def self.download_translated_phrases(filename, project_id:, language_code:)
+    logger.info "Downloading translated phrases for: #{filename} with language: #{language_code}"
+
+    response = RestClient.get "https://platform.api.onesky.io/1/projects/#{project_id}/translations",
+                              params: headers(language_code).merge(
+                                source_file_name: filename, export_file_name: filename
+                              )
+
+    # NOTE: maybe Error::TextNotFoundError should have been a error under a OneSky namespace
+    raise Error::TextNotFoundError, 'No translated phrases found for this language.' if response.code == 204
```
frett: @kares I know this wasn't originally part of the issue/PR and was the original behavior, but I think if OneSky returns no translations we should return an empty hash instead of throwing an error here.

If a phrase that is required is missing, it will throw an error here: https://github.com/CruGlobal/mobile-content-api/pull/177/files#diff-7077e27c99fca731384a65b65766dce2R20

I'll leave it up to you if you want to include that change in this PR or as a small PR after this one is merged

frett: the reasoning behind relaxing this is that if a page is published that doesn't have any OneSky phrases to translate, we currently need to load a placeholder phrase so that we can publish the page. We shouldn't be dependent on having translations in OneSky if we don't actually use them.

frett:
from a high level architecture view it looks good, I'll defer to @Omicron7 for the rest of the review

PR #181, opened 2019-07-23
GT-518: Support language specific manifests

kares:
a custom manifest might be set for a given language and resource
in that sense its unique - only one can exists for given (resource_id, language_id)

builds on top of [(GT-517)](https://github.com/CruGlobal/mobile-content-api/pull/179)

re-arranged routes (to skip some unused ones -> returned 404 will be the same) 
and moved gems under their proper development/test section

*NOTE: includes a commit from #179 (will rebase after its merged)*

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,21 +32,16 @@ gem 'syslog-logger'
 gem 'active_model_serializers', '= 0.10.5'
 gem 'aws-sdk'
 gem 'aws-sdk-rails'
-gem 'codecov', require: false
 gem 'ddtrace'
 gem 'loofah', '~> 2.2.1'
 gem 'nokogiri', '>= 1.8.5'
 gem 'oj', '~> 2.12.14'
 gem 'paperclip'
 gem 'pg'
-gem 'raddocs'
 gem 'rails-html-sanitizer', '>= 1.0.4'
 gem 'rest-client', '~> 2.0.1'
 gem 'rollbar'
-gem 'rspec_api_documentation'
-gem 'rubocop-rspec'
 gem 'rubyzip', '>= 1.2.2'
-gem 'sinatra', '~> 2.0.2'
 gem 'validates_email_format_of'
 
 # Use Capistrano for deployment
@@ -55,16 +50,21 @@ gem 'validates_email_format_of'
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug', platform: :mri
+  gem 'codecov', require: false
   gem 'dotenv-rails'
   gem 'equivalent-xml', '~> 0.6.0'
   gem 'guard-rspec'
   gem 'guard-rubocop'
   gem 'rspec'
   gem 'rspec-rails', '~> 3.4'
+  gem 'rspec_api_documentation', require: false
   gem 'rubocop', '~> 0.49.0', require: false
+  gem 'rubocop-rspec', require: false
   gem 'simplecov', require: false
   gem 'webmock', require: false
 end
+
+gem 'raddocs'
 
 group :development do
   # Access an IRB console on exception pages or by using <%= console %> anywhere in the code.
--- a/Gemfile
+++ b/Gemfile
@@ -57,7 +57,7 @@ group :development, :test do
   gem 'guard-rubocop'
   gem 'rspec'
   gem 'rspec-rails', '~> 3.4'
-  gem 'rspec_api_documentation', require: false
+  gem 'rspec_api_documentation'
   gem 'rubocop', '~> 0.49.0', require: false
   gem 'rubocop-rspec', require: false
   gem 'simplecov', require: false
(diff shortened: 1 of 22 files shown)


spec/acceptance/follow_ups_controller_spec.rb
```diff
@@ -2,14 +2,14 @@
...
-resource 'Follow Ups' do
+resource 'FollowUps' do
   header 'Accept', 'application/vnd.api+json'
   header 'Content-Type', 'application/vnd.api+json'
   let(:raw_post) { params.to_json }
 
   post 'follow_ups/' do
     let(:data) do
-      { type: :follow_up,
+      { type: type,
```
frett: I'm not seeing where `type` is defined like in the other specs? (inexperienced rails developer question)

Gemfile
```diff
@@ -55,16 +50,21 @@ gem 'validates_email_format_of'
...
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug', platform: :mri
+  gem 'codecov', require: false
   gem 'dotenv-rails'
   gem 'equivalent-xml', '~> 0.6.0'
   gem 'guard-rspec'
   gem 'guard-rubocop'
   gem 'rspec'
   gem 'rspec-rails', '~> 3.4'
+  gem 'rspec_api_documentation', require: false
```
Omicron7: We're going to test this in our build system. I think these specific gems were in the production group because our Docker build runs the `docs:generate`. I'll let you know if we need to move these back.
https://github.com/CruGlobal/mobile-content-api/blob/master/Dockerfile#L19

app/controllers/custom_manifests_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+  def show
+    manifest = find_manifest
+    render json: manifest, status: :ok
+  end
+
+  private
+
+  def find_manifest
+    if params[:id]
+      CustomManifest.find(params[:id])
```
Omicron7: Should this be `find!`? I don't know the specific use cases here, but I'd think if `params[:id]` was set, but not found, we wouldn't want the controller to create/update and instead raise an error.

spec/acceptance/follow_ups_controller_spec.rb
kares: `type` is a helper (from the acceptance spec helper) and uses a naming convention: 
https://github.com/CruGlobal/mobile-content-api/pull/181/commits/c554d0cc3aa55e508dda2c5bd3e74bf2abe37412#diff-4d353319af1985fd16aee40f0bbb5e6cR25

... of course any explicit `let(:type) { 'some-name'' }` will override

kares: moved that commit to https://github.com/CruGlobal/mobile-content-api/pull/179 (as I was fixing coverage)
... so once that is merged I will rebase here

frett: Was guessing it was my lack of rails experience :)

frett:
@kares go ahead and rebase, once that is done we will get this into our stage environment for some testing

app/controllers/custom_manifests_controller.rb
kares: this should raise, `find` is noisy when a nil or non existing id string : 
(I also always forget which one raises and which one does go silently)
```
2.3.6 :025 > Translation.find(1)
  Translation Load (0.7ms)  SELECT  "translations".* FROM "translations" WHERE "translations"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
 => #<Translation id: 1, is_published: true, version: 1, resource_id: 1, language_id: 1, translated_name: nil, translated_description: nil, manifest_name: nil, translated_tagline: nil> 
2.3.6 :026 > Translation.find(10)
  Translation Load (0.6ms)  SELECT  "translations".* FROM "translations" WHERE "translations"."id" = $1 LIMIT $2  [["id", 10], ["LIMIT", 1]]
ActiveRecord::RecordNotFound: Couldn't find Translation with 'id'=10
	from /opt/local/rvm/gems/ruby-2.3.6@cru/gems/activerecord-5.0.7.2/lib/active_record/core.rb:173:in `find'

```

Gemfile
kares: I see, thanks for clarifying (it seemed confusing) 
but since that command sets `RAILS_ENV=test` it should do fine
I am probably just being over-cautions here as I've been bitten by rspec previously ... 
and it makes me worried that by any chance it might get loaded on production

kares: restored the implicit require: https://github.com/CruGlobal/mobile-content-api/pull/181/commits/927c8dfaab0b985ef2889eb59cdefd3d9e6144c6

and now `bundle exec rake db:create db:setup docs:generate RAILS_ENV=test` works

thanks for pointing this one out

kares:
@frett nice catch, updated - we should be handling that case now

frett:
This PR is good, there is one more place we need to expose custom manifests. I sent you a message about it, it should hopefully be a fairly trivial change that can be done in a separate PR

PR #182, opened 2019-07-23
[fix] do proper true/false param conversion

kares:
as currently **PUT /resources/1/onesky** *keep-existing-phrases=false*
will end up doing a `push_new_onesky_translation(keep_existing_phrases = true)`

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -67,6 +67,18 @@ class ApplicationController < ActionController::Base
     params.merge!(ActiveSupport::JSON.decode(request.body.string))
   end
 
+  # Get a parameter as a boolean value.
+  # 'false', '0' (or if param is not present) are false-y.
+  # @return [true, false]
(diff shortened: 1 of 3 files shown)


Omicron7:
We tend to use `ActiveModel::Type::Boolean.new.cast(value)` for this.

kares:
Thanks Brian, I tend to not like using `AM::Type` so much for these.
But if you are used to that I am fine with it and will update the PR.

kares:
updated, a problem with the `FALSE_VALUES` approach would be handling of `0` kinds.
currently `0` is false-y but `0.0` isn't ... its minor but something to keep in mind.
in a rare case of having a rest client encode a zero value as `0.0`.

note: that we can not fix the case by simply having a Set with `0.0` added -> as than the raw `'0.0'` would behave differently than `0.0` and since there are infinite ways of a 0.0 in a string `'0.000'` ...

PR #185, opened 2019-07-24
[ci] match Ruby/Postgres version with production

kares:


kares:
the coverage report doesn't make much sense - maybe is should get relaxed if its blocking merging.
99% coverage does not mean much. would take a 70% quality coverage over an arbitrary 95+% one.

kares:
added some coverage for the problematic path, just a bit crazy that it comes off with unrelated changes

PR #191, opened 2019-07-25
upgrade app to run with ruby 2.5

kares:
as 2.3 (even 2.4) are EOL and no longer supported

with the new Ruby version a gem update was needed (oj was using old deprecated native ext API)
\+ rubocop needs to be newer -> means it will bring in more cops on us, and thus the noisy parts ...

also I am not sure about the Docker image, since I can not access the registry.

*NOTE: this PR assumes https://github.com/CruGlobal/mobile-content-api/pull/185 merged first (to avoid conflicts)*


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM 056154071827.dkr.ecr.us-east-1.amazonaws.com/base-image-ruby-version-arg:2.3
+FROM 056154071827.dkr.ecr.us-east-1.amazonaws.com/base-image-ruby-version-arg:2.5
```
kares: @Omicron7 not sure about the Docker image (can not access the registry), is this enough?

frett:
@kares you will need to address the merge conflicts

kares:
@frett yy, was waiting for @Omicron7 to confirm the registry URL

frett:
the registry URL is correct, it fully built and deployed to Docker in AWS

Dockerfile
Omicron7: Yeah, that works.

PR #238, opened 2019-09-24
Fix URL parsing

jbirdjavi:


frett:
it's quite the coverage drop for a small change

PR #275, opened 2019-11-20
[GT-643] Global activity analytics API

ka8725:
Ticket: https://jira.cru.org/browse/GT-643

- [x] Create the request for access token (default expiration 24 hours) in exchange for JWT
- [x] Call Adobe Analytics through HTTP to get the data, use access token, endpoint is configurable
- [x] Create the table to keep necessary data
- [x] Create API controller exposes the data
- [x] Fill in the table with data coming from Adobe Analytics API
- [x] Trigger counters update
- [x] Update README

**Questions:**

- What is the format for access token request in exchange for JWT token?
Answer: 
```
curl -X POST -H "Cache-Control: no-cache" -H "Content-Type: multipart/form-data" -F "client_id=<ADOBE_ANALYTICS_CLIENT_ID>" -F "client_secret=<ADOBE_ANALYTICS_CLIENT_SECRET>" -F "jwt_token=<ADOBE_ANALYTICS_JWT_TOKEN>" "<ADOBE_ANALYTICS_EXCHANGE_JWT_URL>"
```
Expect Response `status` 200 with `body`:
(note that the `expires_in` value will not be correct)
```
{
    "token_type": "bearer",
    "access_token": "lkjasdfkjlhasdflkjhasdflkjhasdfjkhaslkjhalskjdfh",
    "expires_in": 86399995
}
```

- What's the correct header for requests to Adobe Analytics? Is the following correct one?

Answer:
```
Authorization: Bearer <access token in response from request to ADOBE_ANALYTICS_EXCHANGE_JWT_URL>
x-api-key: <ADOBE_ANALYTICS_CLIENT_ID>
x-proxy-global-company-id: <ADOBE_ANALYTICS_COMPANY_ID>
```
- Request format: is the `Client ID` in the header correct?

Answer: No Client ID necessary.

- Response format: 

```
    "rows": [
        {
            "itemId": "1173709922",
            "value": "Some Country",
            "data": [
                123,
                123,
                123,
                218
            ]
        },
```
what the numbers in "data" array mean? `users countries launches gospel_presentations`? is that correct order?
Answer: 1 Unique Visitors, 2 Launches, 3 Presenting the Gospel, 4 Total Countries (row count)
3. Response format: can we take counters from this piece of data rather than reducing manually?

Answer: Yes.

```
    "summaryData": {
        "filteredTotals": [
            238326,
            966442,
            43834,
            218
        ],
        "totals": [
            238326,
            966442,
            43834,
            218
        ],
        "col-max": [
            42828,
            165291,
            11977,
            218
        ],
        "col-min": [
            1,
            1,
            1,
            218
        ]
    }
```

### Testing instructions

1. Define the following env vars:
```
    * ADOBE_ANALYTICS_REPORT_URL
    * ADOBE_ANALYTICS_COMPANY_ID
    * ADOBE_ANALYTICS_CLIENT_ID
    * ADOBE_ANALYTICS_JWT_TOKEN
    * ADOBE_ANALYTICS_CLIENT_SECRET
    * ADOBE_ANALYTICS_EXCHANGE_JWT_URL
```
2. Run rails server: `rails s`
3. Check the endpoint: `curl localhost:3000//analytics/global`


app/models/global_activity_analytics.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class GlobalActivityAnalytics < ApplicationRecord
```
ka8725: add docs

app/controllers/global_activity_analytics_controller.rb
```diff
@@ -0,0 +1,10 @@
+class GlobalActivityAnalyticsController < ApplicationController
+  def show
+    begin
+      UpdateGlobalActivityAnalytics.new.perform
+    rescue => e
+      Rollbar.error(e)
+    end
```
jbirdjavi: I'm pretty sure this `rescue` block is unnecessary since Rollbar will capture all exceptions automatically.

app/lib/update_global_activity_analytics.rb
```diff
@@ -0,0 +1,119 @@
...
+    @analytics_url = ENV.fetch("ADOBE_ANALYTICS_REPORT_URL")
+    @company_id = ENV.fetch("ADOBE_ANALYTICS_COMPANY_ID")
+    @client_id = ENV.fetch("ADOBE_ANALYTICS_CLIENT_ID")
+    @jwt_token = ENV.fetch("ADOBE_ANALYTICS_JWT_TOKEN")
+    @client_secret = ENV.fetch("ADOBE_ANALYTICS_CLIENT_SECRET")
+    @exchange_jwt_url = ENV.fetch("ADOBE_ANALYTICS_EXCHANGE_JWT_URL")
+  end
+
+  def perform
+    return if analytics.actual?
```
jbirdjavi: It'd be nice if all of the instance vars in this class would use `@`, e.g. `@analytics`

db/migrate/20191120152056_add_global_activity_analytics.rb
```diff
@@ -0,0 +1,19 @@
...
+      create table global_activity_analytics (
+        id integer primary key default 1,
+        users integer default 0 not null,
+        countries integer default 0 not null,
+        launches integer default 0 not null,
+        gospel_presentations integer default 0 not null,
+        updated_at timestamp without time zone not null,
+        constraint singleton check (id = 1)
+      )
+    SQL
```
jbirdjavi: Can this be done with ruby code?  It looks like there's a `singleton_guard` option that might work.

app/lib/update_global_activity_analytics.rb
ka8725: I would appreciate explanation why it would be better. But ok, I'm changing this.

app/controllers/global_activity_analytics_controller.rb
ka8725: Answering to the inlined question and this one:
> Is this supposed to automatically refresh every 24 hours? If so, I don't see how that happens.

With current implementation update cache is doing *on demand* but not more than once in 24h. The cache update is triggered right from this controller action. How does it to work:
- on the very first request the cache is outdated (it's `GlobalActivityAnalytics.instance.updated_at` field is later than 24h ago, see `GlobalActivityAnalytics.instance`  implemetation)
- when a mobile client performs the request the two cases are possible:
-- cache is up to date: `GlobalActivityAnalytics.instance.updated_at` <= 24h.
   In this case `UpdateGlobalActivityAnalytics.new.perform` will do nothing (see guard clause in `UpdateGlobalActivityAnalytics#perform` - `return if @analytics.actual?`), i.e. `rescue` won't be evaluated
-- cache is outdated: `GlobalActivityAnalytics.instance.updated_at` > 24h.
   In this case `UpdateGlobalActivityAnalytics.new.perform` will try to update the counters by requesting analytics API. Since it's a network call various exceptions are possible: network connection error, timeouts, TLS, etc.
   I don't want to expose these exceptions to the end client and want to return some data always. But these exceptions is an important information for developers, and most probably they would need to react on them. Therefore, it catches these exceptions and returns the previously cached results to the end-user. To implement the monitoring for developers it logs the exception into Rollbar. It's done with `rescue`

app/lib/update_global_activity_analytics.rb
jbirdjavi: It's much easier to see if the code is referring to an instance variable or a method call that way.

app/controllers/global_activity_analytics_controller.rb
jbirdjavi: Got it.  Thanks.

PR #282, opened 2019-12-02
Output https URLs for attachments

jbirdjavi:


wrandall22:
Do we know if there was a reason to send them as `http` in the first place, or if it was just overlooked?

jbirdjavi:
@wrandall22 I worked on the migration for attachments to use ActiveStorage instead of Paperclip, and the default in ActiveStorage was to use http.  From what I investigated, it looks like it was linking straight to S3 using https before the migration.

PR #335, opened 2020-02-21
GT-698: Allow a GodTools user to sign up for content hosted in Adobe Campaigns (WIP)

ckozus:


.env
```diff
@@ -20,4 +20,13 @@ REDIS_PORT_6379_TCP_PORT=
...
+DEFAULT_PROXY_TARGET=
+
+# Adobe Campaigns information and credentials
+ADOBE_ORG_ID=asdf
+ADOBE_ORG_NAME=cru
+ADOBE_TECH_ACCT=asdf
+ADOBE_API_KEY=asdf
+ADOBE_API_SECRET=asdf
+ADOBE_SIGNED_JWT=asdf
+ADOBE_SERVICE_NAME=GodToolsWelcomeSeries
```
frett: we will need the ability to support multiple different adobe campaign series, so my guess is this probably shouldn't be defined as an environment variable. Instead the `destination` model should track whatever is needed to identify/submit to a specific series.

I don't know the campaigns API, but it might be able to share the API_KEY, API_SECRET, and SIGNED_JWT across all series and it may just be the series/service identifier that needs to be configured per destination.

ckozus: hi @frett the ticket originally said that I should use a specific series. But yes, it shouldn't be a problem to store the series name in a different part, I think the `destinations` table would be the perfect place, so callers could use different `destination_id` depending which series they want.

frett: I'll go update the ticket this afternoon to reflect the desire to support multiple campaigns.

ckozus: @frett done! I have implemented those changes.

db/migrate/20200220150100_add_service_type_to_destination.rb
```diff
@@ -0,0 +1,20 @@
+class AddServiceTypeToDestination < ActiveRecord::Migration[5.2]
+  def up
+    add_column :destinations, :service_type, :string
+    add_column :destinations, :adobe_series_name, :string
```
hlbraddock: Please change `series_name` to `service_name`.

app/models/follow_up.rb
```diff
@@ -42,8 +42,23 @@ def headers
   end
 
   def perform_request
+    case destination.service_type
```
hlbraddock: This service type driven request design is very simple and nicely designed. 👍

.env
```diff
@@ -20,4 +20,12 @@ REDIS_PORT_6379_TCP_PORT=
...
 SIDEKIQ_CREDS=
-DEFAULT_PROXY_TARGET=
\ No newline at end of file
+DEFAULT_PROXY_TARGET=
+
+# Adobe Campaigns information and credentials
+ADOBE_ORG_ID=asdf
+ADOBE_ORG_NAME=cru
+ADOBE_TECH_ACCT=asdf
+ADOBE_API_KEY=asdf
```
hlbraddock: You're still planning on moving these (`ADOBE_API_KEY `, `ADOBE_API_SECRET`) to `destinations` table, correct?

ckozus: @hlbraddock that's correct.

app/models/destination.rb
```diff
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
 class Destination < ActiveRecord::Base
-  validates :url, presence: true
+  enum service_type: {growth_spaces: "growth_spaces", adobe_campaigns: "adobe_campaigns"}
+
+  validates :url, :service_type, presence: true
+  validates :route_id, presence: true, if: :growth_spaces?
```
hlbraddock: I should think the `presence` of a `route_id` is also unconditional. Is there some use case where it might not be available?

ckozus:
Closed in favor of #337 and #338

app/models/destination.rb
ckozus: @hlbraddock `route_id` is always used as a parameter when calling the Growth Spaces API, so I assume it's a mandatory parameter, that's why I added the validation.
Do you think I should remove it?

hlbraddock: When I say that `route_id` is 'unconditional', I mean to say that it is **always** required, not just for `growth_spaces`.

PR #337, opened 2020-02-24
GT-698: [step-1] Allow a GodTools user to sign up for content hosted in Adobe Campaigns

ckozus:
This is the first of two PRs, the first one adds adobe related fields to the destinations table, allowing them to be filled up before deploying further changes (which will require this information).

### DIFF
--- a/db/migrate/20200220150100_add_service_type_to_destination.rb
+++ b/db/migrate/20200220150100_add_service_type_to_destination.rb
@@ -1,0 +1,21 @@
+class AddServiceTypeToDestination < ActiveRecord::Migration[5.2]
+  def up
+    add_column :destinations, :service_type, :string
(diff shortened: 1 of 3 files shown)


db/migrate/20200220150100_add_service_type_to_destination.rb
```diff
@@ -0,0 +1,25 @@
+class AddServiceTypeToDestination < ActiveRecord::Migration[5.2]
+  def up
+    add_column :destinations, :service_type, :string
+    add_column :destinations, :service_name, :string
+    add_column :destinations, :adobe_api_key, :string
```
hlbraddock: We'll not be needing to add `adobe_api_key` or `adobe_api_secret` as the columns `access_key_id` and `access_key_secret` already exist. @ckozus

PR #338, opened 2020-02-24
GT-698: [part-2] API - Allow a GodTools user to sign up for content hosted in Adobe Campaigns

ckozus:

Needs to be deployed after #337 and once the new records on the `destinations` table are completed with credentials from adobe.

This PR follows the pattern of https://github.com/CruGlobal/missionhub-api/pull/215.

Before deploying this, the following env variables need to be defined:

```
# Adobe Campaigns information and credentials
ADOBE_CAMPAIGN_ORG_ID=<value>
ADOBE_CAMPAIGN_ORG_NAME=<value>
ADOBE_CAMPAIGN_TECH_ACCT=<value>
ADOBE_CAMPAIGN_SIGNED_JWT=<value>
```


ckozus:
@frett hey! Are you familiar with the Travis error we're getting on this PR? Doesn't seem related to my changes (I have already tried restarting the build, but with no changes).

config/initializers/adobe.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+# Get credentials from database
+destination = Destination.adobe_campaigns.first
+if destination&.adobe_api_key.present? && destination&.adobe_api_secret.present?
```
hlbraddock: Note that existing columns `access_key_id` and `access_key_secret` should be referenced instead of `adobe_api_key` and `adobe_api_secret`, respectively.

app/models/destination.rb
```diff
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
 class Destination < ActiveRecord::Base
-  validates :url, presence: true
+  enum service_type: {growth_spaces: "growth_spaces", adobe_campaigns: "adobe_campaigns"}
+
+  validates :url, :service_type, presence: true
+  validates :route_id, presence: true, if: :growth_spaces?
```
hlbraddock: The presence of `route_id` should always be required (not just for `growth_spaces`).

ckozus: @hlbraddock what value should it contain for adobe records?

config/initializers/adobe.rb
ckozus: @hlbraddock got it, I didn't get that part originally.

app/models/destination.rb
hlbraddock: Some number (though the actual column data type is `character varying`) just like for growth spaces. It's the way the client specifies the service desired. All services will therefore have route id's.

ckozus: @hlbraddock actually clients specify the service they want using a `destination_id` as a parameter of the request:

https://github.com/CruGlobal/mobile-content-api/blob/master/app/controllers/follow_ups_controller.rb#L13

`route_id` is a growth_space related parameter:

https://www.growthspaces.org/api/docs#subscribers

app/services/adobe_campaign.rb
```diff
@@ -0,0 +1,54 @@
...
+    @adobe_profile ||= post_to_adobe_campaign
+  end
+
+  def find_on_adobe_campaign
+    Adobe::Campaign::Profile.by_email(follow_up.email)["content"][0]
+  end
+
+  def post_to_adobe_campaign
+    Adobe::Campaign::Profile.post("email": follow_up.email,
+                                  "firstName": follow_up.name,
```
hlbraddock: Is it possible to set `firstName` and `lastName` from `names.first_name` and `names.last_name` from https://github.com/CruGlobal/mobile-content-api/blob/master/app/models/follow_up.rb#L36?

ckozus: @hlbraddock done!

config/initializers/adobe.rb
```diff
@@ -20,6 +20,6 @@
   else
     Rollbar.warning("Adobe Campaigns was not initialized, check info on destinations table")
   end
-rescue ::ActiveRecord::NoDatabaseError => e
+rescue ::ActiveRecord::NoDatabaseError
```
hlbraddock: Looks like you resolved cause of your Travis build failure here.

app/models/destination.rb
```diff
@@ -1,5 +1,9 @@
...
+  validates :service_name, presence: true, if: :adobe_campaigns?
 end
```
hlbraddock: I wonder should we add:
`validates :access_key_id, :access_key_secret presence: true`
or 
`validates :access_key_id, :access_key_secret presence: true, if :growth_spaces? or :adobe_campaigns?` (syntax??)

ckozus: @hlbraddock I also think we should add the validation. 
The second option is not really necessary, since those are the only two possible options at the time.

spec/support/fixtures/adobe_campaign_parsed_json_fixtures.rb
```diff
@@ -0,0 +1,180 @@
...
+      "cusJOBTITLE" => "",
+      "cusLink" => {
+        "href" => "https://mc.adobe.io/cru/campaign/profileAndServicesExt/profile/@EJcyhZ11jryZtasdfIAHR2c8foWIzASt5-c9p3FShoQR8nTDjEp7xY55lHFPiEffYnCm8nsIW0zIvoVt0-3zrQ_GjTs/cusLink/",
+      },
+      "cusLink2" => {
+        "href" => "https://mc.adobe.io/cru/campaign/profileAndServicesExt/profile/@EJcyhZ11jryZtasdfIAHR2c8foWIzASt5-c9p3FShoTomQPm9e3ETJ_s2pUrH5DVPZuk_Z0W41Q6y_gWBZNXq0tVd1bKDLzkB92oTy7ZdPgYWNvX/cusLink2/",
+      },
+      "cusMinistryCode" => "",
+      "cusMinistryCodeGR" => "",
+      "cusMissionHubUser" => "Y",
```
hlbraddock: You should probably change all `MissionHub` references to `GodTools` (or `MobileContent`).
@ckozus

ckozus: @hlbraddock done!

hlbraddock:
@jbirdjavi Can you add your review?  Note that this passes QA when run locally. I'll soon be deploying to staging.

hlbraddock:
Note that @ckozus anticipates the code coverage failure (https://codecov.io/gh/CruGlobal/mobile-content-api/compare/5d80daf1af23ac4ab0757c99ce83dfeebe5d86ed...3c973b6576f1cabf805e87ee412cb3bc23165c58) to be resolved once https://github.com/CruGlobal/mobile-content-api/pull/337 is merged to **_master_** and this PR is then modified to merge with **_master_**.

PR #358, opened 2020-03-11
validate the XML schemas with travis

frett:
We currently keep our XML Schemas for GodTools in this repo because they are enforced by the mobile-content-api only.

This PR adds some xmllint tests to ensure changes to the XML schema doesn't break existing tools in unexpected ways.


frett:
@jbirdjavi I'm not sure if there is a more appropriate mechanism for defining these tests since they don't require ruby to run at all. I'm definitely looking for feedback in there is a convention I should follow for this PR.

PR #391, opened 2020-05-08
GT-783, first step implementing websockets for godtools

andrewroth:
the consumer socket stuff is still to do

lograge broke with action cable so removed for now


app/channels/publish_channel.rb
```diff
@@ -0,0 +1,15 @@
+class PublishChannel < ApplicationCable::Channel
+  def subscribed
+    puts("[PublishChannel#subscibed] #{params.inspect}")
+    stream_for params["channelId"]
+    transmit(consumerChannelId: Time.now.to_i)
```
frett, line 5: I know you are planning on preserving this value at some point, I would request this change to being initialized as a random value at that point

andrewroth: Oh, I totally agree -- just want a test value here.  Definitely will be properly random later.

frett:
@knutsenm the separate publish & subscribe channels is a security precaution to prevent subscribers from being able to push events to other subscribers.

Dockerfile
```diff
@@ -25,6 +25,7 @@ ARG TEST_DB_HOST=localhost
 ARG TEST_DB_PORT=5432
 
 RUN bundle exec rake db:create db:setup docs:generate RAILS_ENV=test
+RUN bundle exec rake assets:clobber assets:precompile RAILS_ENV=test
```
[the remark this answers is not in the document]
andrewroth: I just don't want to touch stuff that's been working for them.  So just adding my line is safe.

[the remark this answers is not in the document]
andrewroth: RAILS_ENV=production doesn't work because it needs other env vars.  This is actually all for only the proof of concept /publish and /subscribe pages which we can remove after, then we can remove this as well.

app/channels/publish_channel.rb
```diff
@@ -0,0 +1,56 @@
+class PublishChannel < BaseSharingChannel
+  def subscribed
+    puts("[PublishChannel#subscribed] #{params.inspect}")
+    @publisher_channel_id = params["channelId"] # may have some validation for valid channels later
+    return unless validate_publisher_channel_id_format
+
+    stream_for @publisher_channel_id
```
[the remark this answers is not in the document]
andrewroth: As discussed in meeting, for the record here: stream_for and broadcast_to seem to work fine with string identifiers as well, and using stream_from and broadcast I would be essentially implementing stream_for and broadcast_to anyways.

app/channels/publish_channel.rb
```diff
@@ -34,23 +34,30 @@ def receive(data)
 	protected
 
   def validate_publisher_channel_id_format
-    if !@publisher_channel_id
-      transmit(error: "publisher channel is missing")
+    if @publisher_channel_id.blank?
+      Rails.logger.info("transmit block here")
+      transmit(format_error("Publisher Channel Missing"))
+      false
+    elsif @publisher_channel_id =~ /...../
```
frett, line 41: am I reading this right that you are limiting publisher channel ids to 5 characters?

knutsenm: This regex is not anchored; it merely tests length >= 5. The string length function would be a better way to do this.

frett: thanks mark, I completely missed the lack of anchors :)

andrewroth: Yeah -- also just want to note that I intended this as more of a stub than anything.  We can tweak that regex to whatever format you want to enforce.

location.conf
```diff
@@ -0,0 +1,58 @@
...
+    proxy_set_header X-Request-Start "t=${msec}";
+}
+
+location /cable {
+    proxy_pass http://puma;
+    proxy_http_version 1.1;
+    proxy_set_header Upgrade $http_upgrade;
+    proxy_set_header Connection "upgrade";
+}
+
```
knutsenm, line 55: The net effect is to add these six lines to the file in ecs_config/ecs/base-image-ruby-version-arg

The danger is that when ecs_config is updated, we are likely to miss those updates here. For example, `voke_api` and `watchthinkchat` have both drifted in this fashion.

It may be wiser to add a script to the build that edits these lines into the existing location.conf, which could be as simple as appending them to the end with `cat`. Thoughts, please, @frett @Omicron7 @jbirdjavi ?

Omicron7: I consider concatenation of config files like this a bad practice. If you'd really like to allow apps to override sections, or add pieces here, we should add and `include` directive to the base image, and let apps create additional .conf files that get included.

knutsenm: > we should add an include directive to the base image

That would be nice, so long as it doesn't error when there is no file to include (the default case). Or, I suppose, the include file could be present in ecs_config but empty.

Omicron7: Something like
```
include /usr/local/openresty/nginx/locations/*.conf;
```
Anywhere `location.conf` is included. Then you copy your `cable.conf` to that location in the Dockerfile and nginx will pull it in.
https://github.com/CruGlobal/ecs_config/blob/master/ecs/base-image-openresty/nginx-httpd.conf#L28

app/channels/publish_channel.rb
knutsenm: @frett what restrictions did you have in mind? Something like alphanumeric and '-' characters only, with a minimum length, would be a superset of the UUID pattern. Could go case insensitive as well, as I believe UUIDs are.

knutsenm: My suggestion: `elsif @publisher_channel_id =~ /\A[a-zA-Z0-9-]{5,200}\z/`

app/channels/subscribe_channel.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    stream_for @subscriber_channel_id
+  end
+
+  def unsubscribed
+  end
+
+  protected
+
+  def validate_subscriber_channel_id_format
```
knutsenm, line 28: This is quite similar to `validate_publisher_channel_id_format` - perhaps both could be factored out into the superclass?

app/channels/publish_channel.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    transmit({data: {type: "publisher-info", attributes: {subscriberChannelId: metadata[:subscriber_channel_id]}}})
+  end
+
+  def unsubscribed
+  end
+
+  def receive(data)
+    @publisher_channel_id = params["channelId"]
```
knutsenm, line 25: Validate id format here as well? Couldn't hurt.

app/channels/publish_channel.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def setup_new_pair
+    subscriber_channel_id = new_random_uid
+    set_metadata(:subscriber_channel_id, subscriber_channel_id)
+    set_metadata(:last_used_at, Time.now)
+    remember_subscriber(subscriber_channel_id)
+  end
+
+  def remember_subscriber(subscriber_channel_id)
+    Rails.cache.write(["subscriber_to_publisher", subscriber_channel_id], params["channelId"])
```
knutsenm, line 58: Probably ought to use `@publisher_channel_id` here instead of `params["channelId"]`

app/channels/publish_channel.rb
andrewroth: Good call

app/channels/subscribe_channel.rb
andrewroth: I was thinking since the publisher ids come from @frett's code, but subscriber ids come from our code, we may want different regex's if they're different, but if not I'm good to combine this

andrewroth: or maybe if that is the case I can just add a block for regex

knutsenm: We ought to be able to agree on reasonable and identical regex for both.

location.conf
frett: I like the idea of supporting additional locations through an include. it would let us fully leverage the base config but still provide overrides when necessary

app/channels/subscribe_channel.rb
frett: honestly I don't care too much about the ids following a specific format beyond these requirements:
- needs to be URI safe because the value is passed to receivers via a url param
- needs to have enough entropy to avoid collisions and prevent guessing of the values

config/environments/staging.rb
```diff
@@ -9,5 +9,5 @@
 
 Rails.application.configure do
   config.secret_key_base = ENV["SECRET_KEY_BASE"]
-  config.action_cable.allowed_request_origins = ["https://mobile-content-api-stage.cru.org"]
+  config.action_cable.disable_request_forgery_protection = false
```
frett, line 12: @andrewroth shouldn't this be `true`?

location.conf
andrewroth: @Omicron7 @frett I created https://github.com/CruGlobal/ecs_config/pull/634/files

app/channels/publish_channel.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def setup_new_pair
+    subscriber_channel_id = new_random_uid
+    set_metadata(:subscriber_channel_id, subscriber_channel_id)
+    set_metadata(:last_used_at, Time.now)
+    remember_subscriber(subscriber_channel_id)
+  end
+
+  def remember_subscriber(subscriber_channel_id)
+    Rails.cache.write([SUBSCRIBER_TO_PUBLISHER, subscriber_channel_id], params["channelId"])
```
knutsenm, line 51: Comment was deemed "Outdated" but still applies: probably ought to use `@publisher_channel_id` here instead of `params["channelId"]`

app/channels/base_sharing_channel.rb
```diff
@@ -0,0 +1,41 @@
...
+    inner_hash["detail"] = detail if detail
+
+    {"errors": [inner_hash]}
+  end
+
+  def validate_channel_id_format(channel_id, channel_name)
+    if channel_id.blank?
+      transmit(format_error("#{channel_name.capitalize} Channel Missing"))
+      false
+    elsif /...../.match?(channel_id)
```
knutsenm, line 34: Need a stricter pattern here

andrewroth: Your suggestion of `/\A[a-zA-Z0-9-]{5,200}\z/` was fine with @frett and I

PR #401, opened 2020-05-26
Various test fixes

knutsenm:


spec/spec_helper.rb
```diff
@@ -56,8 +56,8 @@
   #   # is tagged with `:focus`, all examples get run. RSpec also provides
   #   # aliases for `it`, `describe`, and `context` that include `:focus`
   #   # metadata: `fit`, `fdescribe` and `fcontext`, respectively.
-  #   config.filter_run_when_matching :focus
-  #
+  config.filter_run_when_matching :focus
```
andrewroth, line 59: Curious why this change?

knutsenm: So I can `, :focus` tests when developing; it's a common setting

andrewroth: Huh, no didn't know about that.  Looks handy

knutsenm:
I've still got spec/models/attachment_spec.rb failing intermittently. It would be nice to figure that out, too, before merging this.

I think that spec may be mucking with the seed data. Also, I notice that db/seeds.rb has `.find_or_create_by!` for everything except Attachment

PR #435, opened 2020-07-09
GT-840, Tips

andrewroth:
WIP


app/lib/package.rb
```diff
@@ -24,11 +26,11 @@ def push_to_s3
...
 
     build_zip
-    upload
+    #upload
 
-    PageClient.delete_temp_dir(@directory)
+    #PageClient.delete_temp_dir(@directory)
   rescue => e
-    PageClient.delete_temp_dir(@directory)
+    #PageClient.delete_temp_dir(@directory)
```
frett, line 33: you probably didn't want to commit these commented out lines

app/lib/package.rb
```diff
@@ -40,7 +42,8 @@ def build_zip
 
     Zip::File.open("#{@directory}/#{@translation.zip_name}", Zip::File::CREATE) do |zip_file|
       add_pages(zip_file, manifest)
-      add_attachments(zip_file, manifest)
+      add_tips(zip_file, manifest)
+      #add_attachments(zip_file, manifest)
```
frett, line 46: `add_attachments` should still run, guessing you commented it out for local testing

app/lib/package.rb
```diff
@@ -52,6 +55,11 @@ def determine_resources(document)
     nodes.each { |node| @resources << node.content }
   end
 
+  def determine_tips(document)
+    nodes = XmlUtil.xpath_namespace(document, XPATH_TIPS.join("|"))
+    nodes.each { |node| @tips << Tip.find_by(name: node.content) }
```
frett, line 60: what happens if the tip isn't found by this statement? we want to raise an error in that situation

app/lib/package.rb
```diff
@@ -62,17 +70,39 @@ def add_pages(zip_file, manifest)
     end
   end
 
+  def add_tips(zip_file, manifest)
+    @tips.each do |tip|
```
frett, line 74: a tip can be referenced from multiple locations in the xml, right now you aren't filtering unique tips. This behavior is similar to how `@resources` includes unique resources

app/lib/package.rb
```diff
@@ -113,6 +145,6 @@ def upload
     Rails.logger.info("Uploading zip to OneSky for translation with id: #{@translation.id}")
 
     obj = self.class.s3_object(@translation)
-    obj.upload_file("#{@directory}/#{@translation.zip_name}", acl: "public-read")
+    #obj.upload_file("#{@directory}/#{@translation.zip_name}", acl: "public-read")
```
frett, line 148: again, assuming you commented this for local testing

app/models/resource_type.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class ResourceType < ActiveRecord::Base
-  validates :name, presence: true, uniqueness: true
+  #validates :name, presence: true, uniqueness: true
```
frett, line 4: local testing change?

app/models/system.rb
```diff
@@ -3,5 +3,5 @@
 class System < ActiveRecord::Base
   has_many :resources
 
-  validates :name, presence: true, uniqueness: true
+  #validates :name, presence: true, uniqueness: true
```
frett, line 6: local testing?

app/models/translation.rb
```diff
@@ -34,6 +34,17 @@ def translated_page(page_id, strict)
     xml.to_s
   end
 
+  def translated_tip(tip_id, strict)
+    tip = Tip.find(tip_id)
+    phrases = download_translated_phrases(tip.name)
```
frett, line 39: for the onesky filename I want to transform the tip name like this:
```
    phrases = download_translated_phrases("tip_${tip.name}.json")
```

frett: that will give the json extension which plays nicer with the OneSky web UI (mistake we made with page names using `.xml`), and also "namespace" tip translation files to avoid potential filename collisions.

app/views/pages/publish.html
```diff
@@ -0,0 +1,54 @@
+Publish Test Page<br/><br/>
```
frett, line 1: you accidentally included the actioncable publish & subscribe testing pages

app/lib/package.rb
frett: also, does this only find tips within the current resource?

app/lib/one_sky.rb
```diff
@@ -10,15 +10,21 @@ module OneSky
...
+    rescue RestClient::BadRequest
+      no_onesky_project = true
+    end
 
-    if response.code == 204
+    if no_onesky_project || response.code == 204
       logger.info "No translated phrases found for: #{filename} with language: #{language_code}"
       return {}
     end
+
```
frett, line 27: I'm intrigued by this logic, was this only for testing locally?

I do want to improve the onesky translation behavior for download translations, but I haven't fully documented how I would actually want it to behave.

The behavior should throw an error if something references a translation that doesn't actually exist, but not require OneSky if there are no actual translations.

app/models/resource_language.rb
```diff
@@ -0,0 +1,28 @@
...
+    "#{resource.id}_#{language.id}"
+  end
+
+  def self.test
+    rl = ResourceLanguage.new
+    rl.resource = Resource.find(3)
+    rl.language = Language.find(12)
+    puts rl.custom_pages.inspect
+    rl
+  end
```
frett, line 27: this is interesting, is this supposed to be part of the actual model or in a spec somewhere?

andrewroth: ah, sorry, this was my own testing code and will be removed before final product

app/lib/one_sky.rb
andrewroth: It came out of some of my testing, I made a draft of a new language that had no data with it, and it was giving "bad request".  It was hindering my testing so I added this.  I didn't think there were any translations in that language because I had just added a new draft, but maybe there was?  To be safe maybe we should revert this before completion.

app/controllers/custom_tips_controller.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class CustomTipsController < SecureController
+  def create
+    create_custom_tip
+  rescue ActiveRecord::RecordInvalid, ActiveRecord::RecordNotUnique
```
knutsenm, line 6: Why even try `update_custom_tip` on `RecordInvalid`?

andrewroth: This was copied from custom pages, I can remove it

db/migrate/20200716141930_create_language_attributes.rb
```diff
@@ -0,0 +1,14 @@
+class CreateLanguageAttributes < ActiveRecord::Migration[5.2]
+  def change
+    create_table :language_attributes do |t|
+      t.integer :language_id, null: false
+      t.integer :resource_id, null: false
+      t.string :key, null: false
+      t.string :value, null: false
+      t.boolean :is_translatable, default: false
+      t.references :resource, null: false
```
knutsenm, line 9: No `t.timestamps`? Sadly, these are missing from many existing tables, but I thought it was best practice to add them everywhere?

knutsenm:
Travis says you need to update Gemfile.lock with `bundle install`

app/models/resource_language.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  belongs_to :resource
+  belongs_to :language
+  has_many :custom_pages, ->(obj) { joins(:page).where(pages: {resource_id: obj.resource.id}) }, through: :language
+  has_many :custom_tips, ->(obj) { joins(:tip).where(tips: {resource_id: obj.resource.id}) }, through: :language
+  has_many :language_attributes, ->(obj) { where(resource: obj.resource) }, through: :language
+
+  def set_data_attributes!(data_attrs)
+    data_attrs.each_pair do |key, value|
+      attr_name = key.scan(/^attr-(.*)$/).first&.first
```
knutsenm, line 16:
```suggestion
      attr_name = key[/^attr-(.*)$/, 1]
```

app/lib/package.rb
```diff
@@ -40,6 +42,7 @@ def build_zip
```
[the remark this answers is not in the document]
frett: the only place that we should need to call `determine_tips` is when processing pages. The manifest xml doesn't supported referencing a tip for inclusion.

This is slightly different behavior than OneSky `i18n-ids` and resources, both of which can be referenced from the manifest xml.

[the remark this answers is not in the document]
frett: If we ever need to add references to tips from the manifest xml we will revisit this and add the appropriate method call here.

app/lib/package.rb
```diff
@@ -40,6 +42,7 @@ def build_zip
 
     Zip::File.open("#{@directory}/#{@translation.zip_name}", Zip::File::CREATE) do |zip_file|
       add_pages(zip_file, manifest)
+      add_tips(zip_file, manifest)
```
frett, line 45: `add_tips` should be gated on the `include_tips` resource language attribute. we only include the tips xml if the attribute is true

app/lib/package.rb
```diff
@@ -52,6 +55,15 @@ def determine_resources(document)
     nodes.each { |node| @resources << node.content }
   end
 
+  def determine_tips(document)
+    nodes = XmlUtil.xpath_namespace(document, XPATH_TIPS.join("|"))
+    nodes.each do |node| 
+      tip = @translation.resource.tips.find_by(name: node.content)
+      raise ActiveRecord::RecordNotFound, "Tip not found: #{node.content}" unless tip
```
frett, line 62: because `add_tips` should be gated on `include_tips`, you will probably want to just track the tip names in `@tips` and raise missing tips within the `add_tips` logic

app/validators/xml_validator.rb
```diff
@@ -16,8 +16,10 @@ def xsd(record)
 
   def xsd_location(record)
     return "manifest.xsd" if record.is_a?(Resource) || record.is_a?(CustomManifest)
+    return "training.xsd" if record.is_a?(Tip)
     return record.resource.resource_type.dtd_file if record.is_a?(Page) || record.is_a?(TranslatedPage)
     return record.page.resource.resource_type.dtd_file if record.is_a?(CustomPage)
+    return record.tip.resource.resource_type.dtd_file if record.is_a?(CustomTip)
```
frett, line 22: `CustomTip` should use the same xsd as `Tip`

app/controllers/resource_languages_controller.rb
```diff
@@ -0,0 +1,26 @@
+class ResourceLanguagesController < SecureController
+  before_action :get_resource, :get_language, :build_resource_language
+
+  def show
+    render json: @resource_language, include: params[:include], status: :ok
+  end
```
frett, line 6: can we make `show` public, but keep `update` private?

app/controllers/resource_languages_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def update
+    @resource_language.set_data_attributes!(data_attrs)
+    render json: @resource_language, status: :ok
+  end
```
frett, line 11: when I send this request:
```
PUT /resources/:resource_id/languages/:language_id

{
    "data": {
        "type": "resource-language",
        "attributes": {
            "attr-test": true
        }
    }
}
```

it returns this attribute: `"attr-test": "t"`, I'm guessing that is an artifact of how ruby represents true/false. can that scenario be transformed into the string "true" instead of "t"

frett: In addition when I send this request:
```
PUT /resources/:resource_id/languages/:language_id

{
    "data": {
        "type": "resource-language",
        "attributes": {
            "attr-test": false
        }
    }
}
```
it deletes `attr-test` instead of setting the value to "false"

db/migrate/20200716141930_create_language_attributes.rb
andrewroth: Good call, thanks, I had copied this from schema.rb's language table

config/routes.rb
```diff
@@ -7,11 +7,15 @@
...
-  resources :resources
+  resources :resources do
+    resources :languages, controller: :resource_languages, only: [:update, :show]
+  end
   resources :drafts
   resources :translations, only: [:index, :show]
   resources :pages, only: [:create, :update, :show]
+  resources :tips, only: [:create, :update, :show]
   resources :custom_pages, only: [:create, :update, :destroy, :show]
+  resources :custom_tips, only: [:create, :update]
```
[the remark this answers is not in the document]
andrewroth: Good catch

app/lib/package.rb
andrewroth: Right, I did it just a bit differently but am checking the flag now

app/controllers/resource_languages_controller.rb
andrewroth: Good catch, thanks

app/validators/xml_validator.rb
andrewroth: Good catch, thanks

app/controllers/resource_languages_controller.rb
andrewroth: Good catch, I've fixed this and added a test for it

app/controllers/resource_languages_controller.rb
andrewroth: Sure

app/lib/one_sky.rb
frett: I still need to digest this change and the impact it will have. This is mostly a note for myself and should block merging this PR until I have a chance to look into this.

frett: I'm good with this change.

knutsenm:
@andrewroth note, I did a couple maintenance commits, you should git pull

PR #454, opened 2020-08-12
Increase the channel expiration time

frett:
This helps ensure that users don't accidentally get an invalid channel due to timeout.


frett:
I'm putting this on hold until we work out the other shared state issue. This may not be needed once that problem is addressed

PR #455, opened 2020-08-13
add redis config for the Rails cache

frett:
Copied and modified from the missionhub-api config


frett:
I see them in secrets, so I'm going to merge

PR #462, opened 2020-08-28
Upgrade to Rails 6

jbirdjavi:


frett:
I'll defer to @knutsenm for review

jbirdjavi:
Deployed to staging.

app/validators/attachment_validator.rb
```diff
@@ -2,7 +2,9 @@
 
 class AttachmentValidator < ActiveModel::Validator
   def validate(model)
-    return unless Attachment.exists?(sha256: model.generate_sha256, resource: model.resource)
+    file = model.attachment_changes["file"].attachable
+    sha256 = XmlUtil.filename_sha(File.open(file).read)
+    return unless Attachment.exists?(sha256: sha256, resource: model.resource)
```
knutsenm, line 7: I'm not sure, but this seems repetitive of the changes to `generate_sha256`. Could this now be reverted?

db/schema.rb
```diff
@@ -222,6 +222,17 @@
...
+    t.integer "id"
+    t.boolean "is_published"
+    t.integer "version"
+    t.integer "resource_id"
+    t.integer "language_id"
+    t.string "translated_name"
+    t.string "translated_description"
+    t.string "manifest_name"
+  end
+
```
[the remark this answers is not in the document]
jbirdjavi: Good catch.  Fixed.

app/validators/attachment_validator.rb
jbirdjavi: So, I feel like you might be right, but the tests failed when I took it out.  The issue is that in Rails 5, the file would exist in the `file` attribute before a save, and in Rails 6, it no longer does.  I made another refactor to the `Attachment` model thinking it would allow this to be taken out, but tests still fail.  I can look more in depth tomorrow if you want.

knutsenm: > the tests failed when I took it out

When was that, mid-development, or after most commits? Maybe it will work now?

jbirdjavi: When I did my refactor yesterday it failed.  I'll investigate more this afternoon yet.

jbirdjavi: Okay, so the problem was that I didn't remove the extra validator code correctly.  It does work without it if I do it right.  :)

PR #466, opened 2020-09-02
some rspec improvements

soberstadt:
- allow using `--only-failures`
- remove database-cleaner as unnecessary (because we are running our tests in transactions)
- add spring to rspec


dbenton9 (commented):
I'm confused why we're adding `--only-failures` when we've already been able to use them?

I know sometimes when we have issues with a spec its because we need to run `spring stop`. Is adding spring to rspec going to inhibit us from running `spring stop`? Or will it be helping those instances we have issues with spring?

soberstadt:
@dbenton9  `--only-failures` requires `config.example_status_persistence_file_path` to be set (which I did in spec_helper.rb)

soberstadt:
Adding spring to rspec is what you have already experienced, that rspec starts up faster because of spring.

dbenton9 (approved):

PR #469, opened 2020-09-02
Add users and okta login support

soberstadt:
Depends on #465


soberstadt:
@frett this will add okta login support. I decided to add a new param (`okta_id_token`) to the existing auth endpoint which will return a similar JWT but with a user id in the payload. How does that sound?

.env
```diff
@@ -29,3 +29,7 @@ ADOBE_CAMPAIGN_ORG_ID=asdf
 ADOBE_CAMPAIGN_ORG_NAME=cru
 ADOBE_CAMPAIGN_TECH_ACCT=asdf
 ADOBE_CAMPAIGN_SIGNED_JWT=asdf
+
+OKTA_SERVER_URL=https://dev1-signon.okta.com
+OKTA_SERVER_PATH=https://dev1-signon.okta.com
+OKTA_SERVER_AUDIENCE=https://dev1-signon.okta.com
```
frett, line 35: I'm guessing you are using this as a placeholder value?

my basic understanding of the id_token at this point is that the audience should be the client_id that the token is generated for. This client_id should be unique per app/service that uses Okta for Auth.

frett:
@soberstadt I'm fine with re-using the existing auth endpoint, I'd also be fine if you wanted to setup new endpoints for auth providers going forward (e.g. `/auth/okta`).

does it complicate the logic having auth handle multiple authentication mechanisms vs having a new endpoint set up per auth provider?

.env
soberstadt: Yes, these are placeholder values. The code will actually support multiple audiences/client_ids which I think we will need to have the web and apps both use Okta login.

frett: just wanted to be sure since this is the first time I'm digging into how Okta handles auth between a mobile app and API

soberstadt:
@frett I could split them up, but I don't think the controller code is too complicated at this point (I moved the validation logic out to the service classes): https://github.com/CruGlobal/mobile-content-api/pull/469/files#diff-d6abe35e81895b206b18f8081d50019c

app/services/okta.rb
```diff
@@ -0,0 +1,69 @@
...
+    def fetch_jwks
+      path = if ENV["OKTA_SERVER_ID"].present?
+        "/oauth2/#{ENV["OKTA_SERVER_ID"]}/v1/keys"
+      else
+        "/oauth2/v1/keys"
+      end
+      response = get(path)
+      JSON.parse(response.body, symbolize_names: true)
+    end
+  end
```
dbenton9, line 65: I see this method constructing a path, then getting the path, and returning the response parsed by JSON to give us the JWKs. What are the JWKs and what are they for?

spec/services/okta_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    {
+      sub: "qwer",
+      email: "okta@okta.com",
+      first_name: "Okta",
+      lsat_name: "Test"
+    }
+  end
+  let(:jwt_payload) { okta_user_info.merge(exp: 1.minute.from_now.to_i) }
+  let(:jwt_default_payload) { {iss: ENV["OKTA_SERVER_PATH"], aud: ENV["OKTA_SERVER_AUDIENCE"]} }
+  let(:rsa_private) { OpenSSL::PKey::RSA.generate 2048 }
```
dbenton9, line 15: What is the `rsa_private`? I see its used to create the `jwk`.

app/models/auth_token.rb
```diff
@@ -51,7 +53,12 @@ def add_issued_at(payload)
   end
 
   def token
-    self.class.generic_token
+    if @user
+      payload = {user_id: @user.id, exp: expiration.to_i}
+      self.class.encode(payload)
```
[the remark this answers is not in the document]
soberstadt: Yes :)

app/models/user.rb
```diff
@@ -0,0 +1,7 @@
+class User < ApplicationRecord
+  validates :okta_id, uniqueness: true, presence: true
+
+  # while the email needs to be validated case-insensitively, we'll
+  # let Rails pass the insensitive check down to postgres's citext type
+  validates :email, uniqueness: {case_sensitive: true}, presence: true
```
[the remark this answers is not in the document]
soberstadt: by default rails makes an inificent uniqueness check if you use `case_sensitive: false` (`where('LOWER(email) = ?', email.downcase)`), but if you use `case_sensitive: true` it uses the slightly faster `where('email = ?', email)`. This ends up being equivalent behavior on the database because the field is treated as case insensitive. does that make sense? (answering questions like this help me get better at communicating about it 😀)

app/services/okta.rb
soberstadt: https://openid.net/specs/draft-jones-json-web-key-03.html
JWK (JSON web key) is a standard on how to structure public signing keys so clients know how to read them.
We use this public signing key (Okta's key) to confirm that the id_token was definitely made by our okta server.

I don't fully understand the internals public-private signing, but it's somewhat magical. Using a private key only the system with the private key can sign (or encrypt) the data, but anyone with the public key can confirm the paired private signed it.

spec/services/okta_spec.rb
soberstadt: RSA is a type of encryption key. Here we create a new public/private key for each test and then test decoding with that key.

app/models/user.rb
```diff
@@ -0,0 +1,7 @@
+class User < ApplicationRecord
+  validates :okta_id, uniqueness: true, presence: true
```
frett, line 2: do we not get a guid back in the OIDC id_token? we are tracking guids within Okta for users, and my understanding is that is still the preferred mechanism of connecting users via different systems.

app/models/user.rb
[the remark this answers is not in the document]
dbenton9: Yes, you explained this well. I understand now. Calling true on case_sensitive helps the sql query be more efficient, but the database still accomplishes the tasks of comparing insensitive case emails.

app/services/okta.rb
dbenton9: I dont fully comprehend this either but I think I get the gist. The public-private keys are used for signing to make sure the data is encrypted and unencrypted between our server and the intended recipients?

soberstadt: yeah, basically.

app/models/user.rb
soberstadt: We do get ssoguid back, I wasn't sure if that was a requirement for this or what the plan is for guids long term. if we're planning to migrate to okta user ids or something.

Do you think I should save the sso guid during login? are new okta users (outside cru) getting new ssoguids?

frett: new accounts get guids created for them, I think we are staying with guids because it's so integral to every system we have, and there isn't a point in migrating away from them.

the dev1 environment isn't set up to generate guids currently though, mainly because it doesn't have any of our infrastructure attached to it.

frett: I would say we would want to record guids to make it easier to integrate with other systems in the future if needed

soberstadt: If that is the case then I should be able to change to sso guid.

soberstadt: changes made in [`4f309af` (#469)](https://github.com/CruGlobal/mobile-content-api/pull/469/commits/4f309af7dbbba695e56328faa68af6e1de24a453)

dbenton9 (approved):

soberstadt:
I'm backing this out of staging because our stage database doesn't have the `citext` extension enabled :(

soberstadt:
I pulled this from staging again because it is failing to deploy, but I don't really know why.

frett:
I merged in master and reset the staging branch

frett:
@soberstadt I just got this rollbar crash on stage: https://rollbar.com/Cru/mobile-content-api/items/218/

frett:
I think the error was from a misconfigured `OKTA_SERVER_ID` environment variable, I just removed it and will be trying again shortly

app/services/okta.rb
```diff
@@ -0,0 +1,71 @@
...
+      User.find_by(sso_guid: user_info[:sso_guid])
+    end
+
+    def create_user(user_info)
+      User.create!(user_info)
+    end
+
+    def validate_okta_id_token(id_token)
+      payload = transform_jwt_payload JWT.decode(id_token, nil, true, jwt_options).first
+      raise "ID Token does not include sso guid, make sure login scope includes profile" unless payload[:sso_guid]
```
frett, line 28: I'm having trouble getting the `id_token` to actually include the sso guid, or any other profile claims beyond the minimal claims.

from the Okta [documentation](https://developer.okta.com/docs/reference/api/oidc/#scope-dependent-claims-not-always-returned)

> Scope-dependent claims are returned differently depending on the values in response_type and the scopes requested

We are using the recommended SPA authorization code flow, which doesn't appear to be including any custom profile attributes in the `id_token`. I was able to fetch all the attributes including `ssoGuid` using the corresponding `access_token` on the `/userinfo` endpoint

frett: this comment is no longer relevant since Spencer is switching to the `access_token`

PR #475, opened 2020-09-04
GT-933 - Fix custom tip update

soberstadt:
I wrote tests for the controller and got the same error that @frett was experiencing. I then fixed it exactly the way he thought we needed to. (maybe he should take over for me 😉)

See convo in previous PR: https://github.com/CruGlobal/mobile-content-api/pull/435#discussion_r458413836


knutsenm (approved):

frett (approved):
This is good to merge whenever

PR #476, opened 2020-09-04
fix code coverage reporting

soberstadt:
docs: https://github.com/simplecov-ruby/simplecov#filters

This should drop coverage percentage to about 95.83%

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -4,7 +4,7 @@ require "simplecov"
 require_relative "./support/test_helpers"
 require_relative "./support/adobe_campaign_stub_helpers"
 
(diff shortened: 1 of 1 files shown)


spec/rails_helper.rb
```diff
@@ -4,7 +4,12 @@
 require_relative "./support/test_helpers"
 require_relative "./support/adobe_campaign_stub_helpers"
 
-SimpleCov.start
+SimpleCov.start 'rails' do
```
frett, line 7: it looks like the single quotes here breaks standardrb

soberstadt: sigh... yes. fixed.

frett (approved):

PR #554, opened 2020-12-31
GT-1005: use google analytics for global activity

dbenton9:
Google Analytics (GA) will be replacing Adobe Analytics for God Tools starting Jan 1, 2021.

This code replaces Adobe Analytics with GA for the Global Activity Report in God Tools.

Please note on review that while the code is 99% the one remaining thing is we're waiting on an S3 bucket setup by Brian Zoetewey's team that will be used hiding our secret key for GA API service account.


app/lib/update_global_activity_analytics.rb
```diff
@@ -51,61 +21,106 @@ def perform
   private
 
   def fetch_data
```
soberstadt, line 23: Since this method is only used in one place we could probably replace it with calling `google_analytics_report` directly.

app/lib/update_global_activity_analytics.rb
```diff
@@ -51,61 +21,106 @@ def perform
...
+      date_ranges: [date_range],
+      metrics: metrics,
+      dimensions: [countries_dimesion]
+    )
+  end
+
+  def google_analytics_report
+    # Create a new report request
+    request = Google::Apis::AnalyticsreportingV4::GetReportsRequest.new(
+      {report_requests: [sessions_and_users_request, gospel_presentations_request, countries_request]}
```
soberstadt, line 121: I'm surprised we need the `{}` here.
Does this work?
```suggestion
      report_requests: [sessions_and_users_request, gospel_presentations_request, countries_request]
```

soberstadt (changes requested):
This mostly looks good. Two little things and issues with the specs we can work on tomorrow.

soberstadt (approved):

PR #558, opened 2021-01-06
remove most aws gems

soberstadt:
while working on #554 I noticed we have a lot of aws gems in here. I think this is due to the aws sdk v2->v3 upgrade when they split them all out. We only actually need s3. Yay cleaner Gemfiles!


jbirdjavi (approved):

soberstadt:
@frett with this on staging on staging now, do you have anything you would like to see QAed before we deploy this? The first smoke test is boot-ability since the api now needs to fetch a file via S3 sdk.

frett:
@soberstadt just tested out publishing a tool, everything looks good to me

soberstadt:
@frett great, thanks for checking. I really don't see anywhere this would be a problem.

PR #612, opened 2021-02-18
GT-1081 enable strict XML validation

frett:
Previously the API would accept some badly formed XML.

For example, the following XML would be accepted as valid even though it didn't have a closing tag:
```xml
<page xmlns="https://mobile-content-api.cru.org">
```

### DIFF
--- a/app/validators/xml_validator.rb
+++ b/app/validators/xml_validator.rb
@@ -2,9 +2,12 @@
 
 class XmlValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
-    xml_errors = xsd(record).validate(Nokogiri::XML(value))
+    xml = Nokogiri::XML(value) { |config| config.strict }
+    xml_errors = xsd(record).validate(xml)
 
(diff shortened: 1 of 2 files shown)


frett:
I just tested this on stage and it is correctly rejecting badly formed xml

app/validators/xml_validator.rb
```diff
@@ -2,9 +2,12 @@
 
 class XmlValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
-    xml_errors = xsd(record).validate(Nokogiri::XML(value))
+    xml = Nokogiri::XML(value) { |config| config.strict }
+    xml_errors = xsd(record).validate(xml)
```
frett, line 6: should I keep this as a single statement? I originally split it into two when I was going to target the rescue to the xml parsing only, and not to the XSD validation

soberstadt: No, I like this better as two lines.

soberstadt (approved):

PR #649, opened 2021-04-06
implement a uniqueness validation on page by resource

andrewroth:


app/models/page.rb
```diff
@@ -7,5 +7,6 @@ class Page < AbstractPage
   validates :filename, presence: true
   validates :resource, presence: true
   validates :position, presence: true, uniqueness: {scope: :resource}
+  validates :filename, presence: true, uniqueness: {scope: :resource}
```
knutsenm, line 10: Should delete line 7

spec/models/page_spec.rb
```diff
@@ -37,6 +37,12 @@
     expect(result.errors["position"]).to include("has already been taken")
   end
 
+  it "cannot duplicate Resource ID and Filename" do
+    result = described_class.create(filename: "04_ThirdPoint.xml", resource_id: 1, structure: structure, position: 1)
+
+    expect(result.errors["filename"]).to include("has already been taken")
```
knutsenm, line 43: OK, but this also breaks the `position` validator, doesn't it?

knutsenm (changes requested):

app/models/page.rb
andrewroth: Thanks, I just pushed a fix. I also reordered the validations so the two scope ones are together, I think that looks a bit nicer

spec/models/page_spec.rb
andrewroth: True, doesn't affect the test, but I fixed it anyhow to make sure we're targeting only filename

knutsenm (approved):

frett (approved):
Tried it out on stage and everything is working correctly

frett:
@andrewroth I'm going to merge this in because it looks good to go

PR #650, opened 2021-04-07
upgrade to ruby 2.7.3 and rails 6.0.3.6

frett:


frett:
Unfortunately Travis doesn't support Ruby 2.6.7 or 2.7.3 yet, until then travis will remain broke. I only updated values and ran appropriate upgrade commands, I haven't actually tested this out yet.

jbirdjavi (approved):
I don't suspect there will be any issues with these updates, so I'm approving.  Up to you if you want to wait for Travis.

PR #689, opened 2021-05-28
GT-1200 Resource language uniqueness

frett:
resource-language attributes should be unique in the context of a `{resource}-{language}-{key}` composite key.

Before this change we couldn't use the same resource-language key on multiple languages for a resource.

### DIFF
--- a/app/models/language_attribute.rb
+++ b/app/models/language_attribute.rb
@@ -3,5 +3,5 @@
 class LanguageAttribute < BaseAttribute
   belongs_to :language
 
-  validates :resource, presence: true, uniqueness: {scope: [:resource, :key]}
+  validates :resource, presence: true, uniqueness: {scope: [:resource, :language, :key]}
 end
--- a/app/models/language_attribute.rb
+++ b/app/models/language_attribute.rb
@@ -3,5 +3,5 @@
 class LanguageAttribute < BaseAttribute
   belongs_to :language
 
-  validates :resource, presence: true, uniqueness: {scope: [:resource, :language, :key]}
+  validates :resource, presence: true, uniqueness: {scope: [:language, :key]}
 end
--- a/spec/acceptance/resource_languages_controller_spec.rb
+++ b/spec/acceptance/resource_languages_controller_spec.rb
@@ -139,8 +139,6 @@ resource "ResourceLanguage" do
         }
       }
     end
-    let(:language2) { Language.second }
-    let!(:attribute) { FactoryBot.create(:language_attribute, language: language2, resource: resource, key: "enable_tips", value: "false") }
 
     requires_authorization
 
--- a/spec/acceptance/resource_languages_controller_spec.rb
+++ b/spec/acceptance/resource_languages_controller_spec.rb
@@ -139,6 +139,8 @@ resource "ResourceLanguage" do
         }
       }
     end
+    let(:language2) { Language.second }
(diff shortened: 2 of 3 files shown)


frett:
I very easily could have missed something or not followed best practices for rails tests, so any feedback is welcome

app/models/language_attribute.rb
```diff
@@ -3,5 +3,5 @@
 class LanguageAttribute < BaseAttribute
   belongs_to :language
 
-  validates :resource, presence: true, uniqueness: {scope: [:resource, :key]}
+  validates :resource, presence: true, uniqueness: {scope: [:resource, :language, :key]}
```
soberstadt, line 6: I know you're not adding it here, but seeing this made me question why the scope is on `:resource` when that is the relationship. With your change, now this validation is doing a query like:
```sql
SELECT 1 AS one 
FROM "language_attributes" 
WHERE "language_attributes"."resource_id" = $1 
AND "language_attributes"."resource_id" = $2 
AND "language_attributes"."language_id" IS NULL 
AND "language_attributes"."key" IS NULL LIMIT $3
[["resource_id", 1], ["resource_id", 1], ["LIMIT", 1]]
```
which obviously is redundant. I think you should clean it up while you're in here:
```suggestion
  validates :resource, presence: true, uniqueness: {scope: [:language, :key]}
```

spec/acceptance/resource_languages_controller_spec.rb
```diff
@@ -139,6 +139,8 @@
         }
       }
     end
+    let(:language2) { Language.second }
+    let!(:attribute) { FactoryBot.create(:language_attribute, language: language2, resource: resource, key: "enable_tips", value: "false") }
```
soberstadt, line 143: New data in an acceptance (integration) spec isn't really the best way to test a validation. For example, if I came back to this spec in a different context, I might be tempted to remove these lines as they don't seem to be doing anything. I think it would be nice to write something out in the LanguageAttribute model spec if we want to test something. But in general I don't usually write/check-in specs for framework tools like simple validations. I would approve this PR without a spec change, or help write a simple model test if you would be more comfortable. I'll defer to you, @frett.

soberstadt (changes requested):

spec/acceptance/resource_languages_controller_spec.rb
frett: k, let me see if I can piece together a model spec

spec/models/language_attribute_spec.rb
```diff
@@ -1,5 +1,21 @@
...
+  let!(:attribute) { FactoryBot.create(:language_attribute, language: language, resource: resource, key: "enable_tips", value: "true") }
+
+  it "resource/language/key combination must be unique and is not case sensitive" do
+    attr = described_class.create(resource_id: resource.id, language_id: language.id, key: "enABle_tips", value: "false")
+
+    expect(attr.errors[:resource]).to include "has already been taken"
+  end
+
+  it "key can be reused by multiple languages for a resource" do
+    attr = described_class.create(resource_id: resource.id, language_id: language2.id, key: "enable_tips", value: "false")
```
soberstadt, line 17: Now I'm really getting nit-picky :P You can save a commit to the database on this one by initializing instead of creating. 
```suggestion
    attr = described_class.new(resource_id: resource.id, language_id: language2.id, key: "enable_tips", value: "false")
```

soberstadt (changes requested):
Great job on the spec :) One little nit-pick.

soberstadt (approved):
thanks for putting up with me :D

PR #693, opened 2021-06-02
cache total views sum to db column

soberstadt:
This addresses https://jira.cru.org/browse/GT-1198

The problem, I believe, is that the `views` table has a lot of rows (2M+) so summing there (even indexed) is slow. Take this recent DD Trace for example:
image
The trace took a total of 6.3s and 85+% of that is spent asking the database for sums.

This PR introduces the [`counter_culture`](https://github.com/magnusvk/counter_culture) gem, which helps us with counters. The idea here being that we can have the database do a "delta" increase to the counter column each time a new view is created. (see demo of counters below).

After @frett made some comments on this PR, I think we can probably use this as a temporary solution and move to doing the deltas ourselves without the views table since that table doesn't have any timestamps or other metadata.

There is some risk to this approach:
- This will slow down the view creation endpoint instead
  - currently the ViewsController#create action is very fast ([95% or requests finished within 65ms](https://app.datadoghq.com/apm/resource/mobile-content-api/rack.request/c6cc84b1f8860e33?end=1622668889197&env=production&index=apm-search&maxPercentile=95&paused=false&query=env%3Aproduction%20service%3Amobile-content-api%20operation_name%3Arack.request%20resource_name%3AViewsController%23create&start=1622064089197))
  - thankfully `counter_culture` does a delta update on the cache column instead of re-summing each time, so I think that should be very fast, only slowing that action down a little. 
- The count cache could drift out-of-sync
  - these values are helpful but not mission-critical
- `counter_culture` dependency cost

This is a multi-stage deployment task:
- add column and start updating value (this PR)
- backfill values (running `View.counter_culture_fix_counts` on prod-console will take care of this)
- remove current `total_views` method from model so serializer starts using this column (follow-up PR 1)
- drop `counter_culture` and make the model do the delta directly instead of saving it to the database (follow-up PR 2)

Sum change demo:
Screen Shot 2021-06-02 at 5 09 17 PM


frett:
I didn't realize the views count was being kept in the database like this. I'm all for consolidating this down to a single field. Increasing time on the create views endpoint is not a big deal.

frett:
as long as we have a per-resource tracking of total views for that resource that supports delta updates, we don't need to keep the individual views breakdown table that we currently have.

frett (approved):
Looks good to me as a non-rails guy

jbirdjavi (approved):

PR #697, opened 2021-06-03
copy the sparse fieldset support from MPDX for use by serializers

frett:


app/controllers/concerns/fields.rb
```diff
@@ -0,0 +1,11 @@
+module Fields
+  private
+
+  def field_params
+    return {} unless params[:fields]
+
+    params[:fields].transform_values do |field_value|
+      field_value.split(",")
+    end.to_unsafe_h
```
soberstadt, line 9: This might be more of a code style thing, but it makes more sense to me to call `to_unsafe_h` on the `params` response and then transform it.

```suggestion
    params[:fields].to_unsafe_h.transform_values do |field_value|
      field_value.split(",")
    end
```

soberstadt (approved):
Looks pretty good to me

app/controllers/concerns/fields.rb
frett: it's fun how code style evolves over time, the positioning of `to_unsafe_h` is the same as in MPDX where you committed it 3 years ago ;)

https://github.com/CruGlobal/mpdx_api/blame/master/app/controllers/concerns/fields.rb#L9

soberstadt: I blame Tatai for not reviewing that massive PR more closely ;)

spec/acceptance/resources_controller_spec.rb
```diff
@@ -42,6 +42,19 @@
...
       expect(JSON.parse(response_body)["included"].count).to be(9)
     end
+
+    it "only get name and system of resources" do
+      do_request 'fields[resource]': "name,system"
+
+      expect(status).to be(200)
+      data = JSON.parse(response_body)["data"][0]
+      attrs = data["attributes"]
+      # expect(attrs.size).to be(1)
```
soberstadt, line 52:
```suggestion
```

soberstadt (changes requested):

spec/acceptance/resources_controller_spec.rb
frett: adjusted the test :)

the fields param doesn't currently apply to our custom attributes, so the first resource returned actually has 3 attributes because of custom attributes.

I switched to testing the second resource returned that doesn't have custom attributes

spec/acceptance/resources_controller_spec.rb
```diff
@@ -42,6 +42,21 @@
...
+      expect(status).to be(200)
+      data = JSON.parse(response_body)["data"][1]
+      attrs = data["attributes"]
+      expect(attrs.size).to be(1)
+      expect(attrs["name"]).not_to be_nil
+      expect(attrs["description"]).to be_nil
+      relationships = data["relationships"]
+      expect(relationships.size).to be(1)
+      expect(relationships["system"]).not_to be_nil
+      expect(relationships["translations"]).to be_nil
```
soberstadt, line 58: if you're just checking keys, you could do this:
```suggestion
      expect(attrs.keys).to eq ["name"]
      relationships = data["relationships"]
      expect(relationships.keys).to eq ["system"]
```

frett: thanks, thats much more succinct, going to make one more adjustment

soberstadt (approved):

PR #699, opened 2021-06-04
add caching to ResourcesController#index

soberstadt:
After #693 and #694, latency dropped down to much more predictable and stable levels:
image

BUT, the latency dropped down to being stable right around 3 seconds. This prompted me to dig in a little more, and I think that we can get away with caching the entire index JSON since it generates the same (very large) json every time until a resource is changed.

One unexpected change needed for this is to add timestamp columns to the `resources` table. It looks like most of the tables here don't have timestamps. I don't plan on doing any backfilling of that data, the cache key will be busted as soon as one of the resources gains an updated at.

One fun thing I learned about during this process is the `cache_key_with_version` a method that uses a hash of the query, the number of results, and the maximum updated_at to build a cache key for an ActiveRecord::Relation. Very cool.


app/controllers/resources_controller.rb
```diff
@@ -36,6 +36,18 @@ def push_to_onesky
 
   private
 
+  def cached_index_json
+    cache_key = Resource.index_cache_key(all_resources, include: params[:include])
```
frett, line 40: needs to include the new sparse fieldset params

app/controllers/resources_controller.rb
```diff
@@ -36,6 +36,18 @@ def push_to_onesky
...
+    Rails.cache.fetch(cache_key, expires_in: 1.hour) { index_json }
+  end
+
+  def index_json
+    ActiveModelSerializers::SerializableResource.new(
+      all_resources.order(name: :asc),
+      include: params[:include]
```
frett, line 47: need the sparse fieldset param here as well

jbirdjavi (commented):
Looks good from a Rails perspective

app/controllers/resources_controller.rb
frett: you also should probably cache on any specified filters (assuming we are actually applying filters)

frett: anyway to serialize all of `params` into a string to avoid needing to account for every individual param in the cache key?

frett (changes requested):
the cache key isn't taking all possible params into account, which could lead to cache corruption between similar but different requests

soberstadt:
@frett were you referring to the fields param you added in #697 or something else?

soberstadt:
For reference, I was getting request times around 900-800ms on staging.

soberstadt:
That went down to 100-70ms with caching!

frett:
> @frett were you referring to the fields param you added in #697 or something else?

yeah, that's what I was referring to, it's also what led to the merge conflict that you just fixed ;)

app/controllers/resources_controller.rb
```diff
@@ -36,6 +36,21 @@ def push_to_onesky
...
+  def index_json
+    ActiveModelSerializers::SerializableResource.new(
+      all_resources.order(name: :asc),
+      include: params[:include],
+      fields: field_params
+    ).to_json
+  end
+
   def all_resources
     if params["filter"]
```
frett, line 55: `filter` params are not currently being observed in the `cache_key`, right now this doesn't matter too much because everything is in the same `system` that we would filter on. But we have had discussions around leveraging `system` to provide different collections of tools, which would mean similar requests with different filter params.

soberstadt: it actually subtly is. because I am sending the result of this method into the cache key calculation and calling `cache_key_with_version` on it. Under the hood eventually rails calls `to_sql` on the relation and hashed the resulting sql for the cache key: https://github.com/rails/rails/blob/v6.0.3.7/activerecord/lib/active_record/relation.rb#L317

frett: nice, didn't realize this (even though you mentioned it in your PR description...)

frett (approved):

frett:
I QAed making a few changes in stage and seeing how it affected API requests, everything appears to be working correctly

app/controllers/resources_controller.rb
soberstadt: maybe you missed it, or maybe you just didn't believe me ;)

soberstadt:
I plan to deploy this Monday morning.

app/controllers/resources_controller.rb
jbirdjavi: It's often difficult to believe @soberstadt, but most of the time, he is correct. ;)

PR #745, opened 2021-08-04
GT-1232 add support for `invisible-if` and `hidden-if` attributes on content elements

frett:
This changes allows elements to dynamically be visible in a UI.

examples:
```xml
<content:button invisible-if="values(quiz1) >= 1" />
<content:button hidden-if="!isSet(quiz1)" />
```

### DIFF
--- a/public/xmlns/content.xsd
+++ b/public/xmlns/content.xsd
@@ -14,6 +14,15 @@
         </xs:restriction>
     </xs:simpleType>
 
+    <xs:simpleType name="expression" final="#all">
+        <xs:annotation>
+            <xs:documentation>This is an expression that evaluates to true or false. The rules for what are valid are
+                defined by the ANTLR4 grammar StateExpression.g4
+            </xs:documentation>
+        </xs:annotation>
+        <xs:restriction base="xs:string" />
+    </xs:simpleType>
+
     <xs:simpleType name="gravity">
         <xs:annotation>
             <xs:documentation>This type defines common gravity constants that determine how elements are positioned.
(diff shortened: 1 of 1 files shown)


levieggertcru:
@frett what would the difference between ```hidden-if``` and ```invisible-if```

frett:
> @frett what would the difference between `hidden-if` and `invisible-if`

@levieggertcru I was trying to capture this android concept: https://developer.android.com/reference/android/view/View#attr_android:visibility
I'm not sure if there is a corresponding concept on iOS, if there isn't we can adjust to find a better solution.

Basically I want to be able to hide elements in 2 different ways. Both mechanisms makes the view not visible. But one of the mechanisms leaves the view as part of the sizing/layout calculations, and the other mechanism completely removes it so it doesn't affect layout of other elements.

The `invisible-if` mechanism would allow the layout to stay stable as the element transitions from visible to not visible or vice-versa.
The `hidden-if` mechanism would cause layout positions to potentially jump when a view passes from visible to not visible or vice-versa.

levieggertcru:
Hey @frett that will work.  iOS will take a little more code because we're using auto layout.  Setting an element as hidden will still leave space for that element until constraints are changed or removed to take away that space.

levieggertcru (approved):

frett:
> Hey @frett that will work. iOS will take a little more code because we're using auto layout. Setting an element as hidden will still leave space for that element until constraints are changed or removed to take away that space.

so that would correspond to the `invisible-if` expression.

`hidden-if` would be that coupled with removing the element from auto-layout as well

frett:
psuedo-code would be:
```
view.hidden = hiddenIf || invisibleIf
if(hiddenIf) {
  delete view.constraints
} else {
  set normal constraints
}
```

levieggertcru:
@frett 
psuedo-code looks correct.
invisible-if we would set constraints in order to take up space for that element.
hidden-if we would not set any constraints for that element.

frett:
@levieggertcru I'm going to rename `hidden-if` to `gone-if`, since it matches the name on android and matches the concept on iOS of completely removing layout constraints

levieggertcru:
Hey @frett that sounds good.

PR #756, opened 2021-10-07
Create GHA workflow for validating XML schemas

frett:
Transition the XML schema validation check from travis to GHA

### DIFF
--- a/.github/workflows/xml-schema-validation.yml
+++ b/.github/workflows/xml-schema-validation.yml
@@ -9,46 +9,29 @@ on:
 
 jobs:
   manifest:
-    name: Manifest XML Schema
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


frett:
I'm going to utilize re-usable workflows to avoid code duplication in the workflow

.github/workflows/xml-schema-validation.yml
```diff
@@ -0,0 +1,37 @@
...
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  workflow_dispatch:
+
+jobs:
+  manifest:
+    name: Manifest
+    uses: CruGlobal/mobile-content-api/.github/workflows/validate-schema.yml@ghaXmlSchemaValidation
```
frett, line 13: I need to change this to reference `master` once this PR actually gets merged.

If I change it before I merge the PR, then it won't be able to find the referenced workflow

PR #806, opened 2021-11-15
GT-1313, implement user counters storing and updating

andrewroth:


app/controllers/application_controller.rb
```diff
@@ -44,17 +44,27 @@ def permit_params(*params)
...
   def authorize!
-    authorization = AuthToken.decode(request.headers["Authorization"])
-
-    if authorization
-      user_id = authorization.first.with_indifferent_access[:user_id]
-      return unless user_id # requested is authorized if using generic JWT
-      return if User.find_by(id: user_id)&.admin
-    end
+    # requested is authorized if using generic JWT (current_user_id will be nil)
+    return if authorization && (current_user_id.nil? || current_user&.admin)
```
frett, line 48: I'm intrigued by this returning authorized if `current_user_id` is nil. To me that seems like a potential security loophole, in what circumstances would the JWT have a nil user id?

I do see the check from the previous version of the code, I'm just curious why this functionality exists.

db/schema.rb
```diff
@@ -239,22 +260,4 @@
...
-  add_foreign_key "follow_ups", "languages"
-  add_foreign_key "pages", "resources"
-  add_foreign_key "resources", "resource_types"
-  add_foreign_key "resources", "systems"
-  add_foreign_key "translated_attributes", "attributes"
-  add_foreign_key "translated_attributes", "translations"
-  add_foreign_key "translated_pages", "languages"
-  add_foreign_key "translated_pages", "resources"
-  add_foreign_key "translations", "languages"
-  add_foreign_key "translations", "resources"
```
frett, line 259: any reason why all the previous foreign keys are being removed?

app/controllers/application_controller.rb
andrewroth: No idea. Just keeping things unchanged, assuming there was some good reason for it.

db/schema.rb
andrewroth: Oh, that's my local db, I'll revert that part

spec/acceptance/translations_controller_spec.rb
```diff
@@ -11,6 +11,7 @@
       do_request
 
       expect(status).to be(200)
+      byebug
```
frett, line 14: you probably didn't mean to commit this

app/controllers/application_controller.rb
frett: Spencer was the one who added it, it might have been related to bridging the old auth mechanism to the new one that actually has users. If that is the case we can drop this scenario.

frett: it looks like an AuthToken can be created without a user here: https://github.com/CruGlobal/mobile-content-api/pull/469/files#diff-dfbe97f70beada492206e82fea7ed6fe79afdf33449579101b1d44effbe8108dR18-R24

this is the old authentication method that we can actually remove. [Here](https://jira.cru.org/browse/GT-1348) is a jira ticket I created to track removing the auth mechanism.

frett: For this PR let's remove the `current_user_id.nil?` condition, the user should only be authorized for admin resources if they are an actual user and are an admin.

knutsenm (changes requested):

db/migrate/20211110153757_create_user_counters.rb
```diff
@@ -0,0 +1,13 @@
+class CreateUserCounters < ActiveRecord::Migration[6.0]
+  def change
+    create_table :user_counters do |t|
+      t.integer :user_id
+      t.string :counter_name
```
frett, line 5: should there be a unique constraint on the combination of these 2 fields? I see you are using `first_or_initialize` when accessing a counter, but it might be beneficial to enforce the constraint at the db level.

andrewroth: Sure, that's a good idea

app/controllers/user_counters_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  protected
+
+  def permitted_params
+    params.require(:data).require(:attributes).permit(:increment)
+  end
+
+  def counter_name
+    counter_name = params[:id]
+    counter_name += ".#{params[:format]}" if params[:format].present?
```
frett, line 19: just curious what the `:format` functionality is intended for?

frett:
I see that you are storing `decayed_count` and `last_decay` in the database, but have no logic implemented around updating those properties yet.

Is that going to be in a subsequent PR?

andrewroth:
Ah right it will go in this one, just haven't gotten to it yet

knutsenm (approved):
Approved once you address @frett review.

frett (changes requested):
still needs the decayed_count functionality implemented

app/controllers/application_controller.rb
```diff
@@ -44,17 +44,27 @@ def permit_params(*params)
...
 
     render_unauthorized
   end
 
+  def authorization
+    @authorization ||= AuthToken.decode(request.headers["Authorization"])
+  end
+
+  def current_user_id
+    @user_id ||= authorization.is_a?(Array) && authorization.first.is_a?(Hash) && authorization.first.with_indifferent_access[:user_id]
```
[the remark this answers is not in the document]
andrewroth: I like having the check for the right structure, but I'm ok with that if you have a strong preference

spec/acceptance/user_counters_controller_spec.rb
```diff
@@ -43,9 +43,25 @@
         expect(JSON.parse(response_body)["data"]).not_to be_nil
         expect(UserCounter.last.count).to eq(70)
         # get close to 45 -- original value of 50 should decay to 25 with the 90 day half-life, then +20 from the patch count incremement
-        expect((UserCounter.last.decayed_count - 45).abs).to be <= 0.004
+        expect((UserCounter.last.decayed_count - 45).abs).to be <= 0.004 # look within 0.004, close enough
```
frett, line 46: 👍

spec/factories/user_counters.rb
```diff
@@ -0,0 +1,9 @@
+FactoryBot.define do
+  factory :user_counter do
+    user_id { 1 }
+    counter_name { "MyString" }
+    count { 1 }
+    decayed_count { 1.5 }
```
frett, line 6: in reality you should never have a `decayed_count` greater than the corresponding `count`. But not that it really matters for any tests.

frett (commented):
I just have the outstanding question on what the `format` param is used for in `counter_name`, otherwise everything looks good

frett:
Screen Shot 2021-11-22 at 3 12 30 PM
when incrementing a new counter, the `last-decay` attribute is being returned as null.

app/controllers/user_counters_controller.rb
andrewroth: rails parses it out of `params[:id]` so I have to add it back in if it's present

frett (approved):

knutsenm:
Travis builds are now disabled in favor of GitHub Actions. When you merge master to this branch it will pick up those changes.

PR #820, opened 2021-12-01
[no-jira] Attempt switch from Travis to GitHub Actions

knutsenm:


knutsenm:
I will adjust the branch production rules after code review.

jbirdjavi (approved):

.github/workflows/ruby.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Check out files
+        uses: actions/checkout@v2
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+
+      - name: Bundle audit
+        run: bundle exec bundle audit check --update --ignore CVE-2015-9284
```
frett, line 61: Any reason we are ignoring a CVE here? I didn't see it ignored in the Travis config

frett (commented):
Just the question about the CVE

.github/workflows/ruby.yml
knutsenm: That is copied over from other Travis configs, for uniformity, but it only applies to OmniAuth and so doesn't matter here.

frett (approved):

knutsenm:
Note that the bash uploader underlying the `codecov` gem is deprecated and going away early next year. We should replace it with v2 of https://github.com/marketplace/actions/codecov

PR #825, opened 2021-12-03
fix test failing with expired access code

andrewroth:


andrewroth:
@frett to my surprise, production was giving errors on tests

frett (approved):

PR #826, opened 2021-12-08
GT-1314, implement saving user count values

andrewroth:


app/serializers/user_counter_serializer.rb
```diff
@@ -3,8 +3,13 @@
 class UserCounterSerializer < ActiveModel::Serializer
   type "user-counter"
   attributes :id, :count, :decayed_count, :last_decay
+  attribute :values, if: :has_values?
```
frett, line 6: we have no need to return the values in the json response to the client. this should save some DB queries on fetching user counters

andrewroth: Hmm what about on the GET requests?

frett: nope, no need for it on GET requests.

db/schema.rb
```diff
@@ -226,12 +226,20 @@
     t.index ["resource_id"], name: "index_translations_on_resource_id"
   end
 
+  create_table "user_counter_values", force: :cascade do |t|
+    t.string "user_id"
```
frett, line 230: should this be `t.integer`?

frett: potentially `user_counter_id` has the same issue

andrewroth: Oh, good catch! Yes

frett: does this same thing apply to `user_counter_id`?

db/migrate/20211201161437_create_user_counter_values.rb
```diff
@@ -0,0 +1,11 @@
+class CreateUserCounterValues < ActiveRecord::Migration[6.0]
+  def change
+    create_table :user_counter_values do |t|
+      t.integer :user_id
+      t.integer :user_counter_id
```
knutsenm, line 5: `t.belongs_to` or `t.references`, perhaps?

app/controllers/user_counters_controller.rb
```diff
@@ -21,12 +23,23 @@ def index
...
   def counter_name
     counter_name = params[:id]
     counter_name += ".#{params[:format]}" if params[:format].present?
     counter_name
   end
+
+  # todo: move to model
+  def apply_values(counter, values)
+    values_before = counter.values
+    new_values = values - values_before
```
knutsenm, line 38: Array subtraction ensures that all `values` will be unique, which you want. However, when I read GT-1314 it also mentions

> We discard any new values that would come before the first current value when alpha sorting the list

and

> We then delete any values that come before a 'deleteBefore' value passed to the API, keeping at least 1 value

I don't see those things done here.

andrewroth: @frett I believe we decided that those requirements weren't actually required for this? Should the JIRA task be updated?

frett: correct, we relaxed that requirement for the time being to see how this is actually used first. There still hasn't been any mobile app work lined up to use the values functionality.

db/migrate/20211201161437_create_user_counter_values.rb
```diff
@@ -0,0 +1,11 @@
...
+      t.string :value
+
+      t.timestamps
+    end
```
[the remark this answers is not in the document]
andrewroth: Ok, I've switched it to an array

spec/acceptance/user_counters_controller_spec.rb
```diff
@@ -73,5 +116,19 @@
...
+    context "with values" do
+      it "includes values" do
+        user_counter.update(values: ["v1", "v2"])
+        do_request
+
+        expect(status).to eq(200)
+        today = Date.today.to_s
+        expected_result = %({"data":[{"id":"tool_opens.kgp","type":"user-counter","attributes":{"count":50,"decayed-count":25.00367978478838,"last-decay":"#{today}"}},{"id":"other.kgp","type":"user-counter","attributes":{"count":60,"decayed-count":20.002943827830705,"last-decay":"#{today}"}}]})
+        puts "expected: #{expected_result}"
+        puts "got:      #{response_body}"
```
knutsenm, line 129: Debugging code, should go away

db/schema.rb
```diff
@@ -233,15 +233,27 @@
...
+    t.integer "id"
+    t.boolean "is_published"
+    t.integer "version"
+    t.integer "resource_id"
+    t.integer "language_id"
+    t.string "translated_name"
+    t.string "translated_description"
+    t.string "manifest_name"
+  end
+
```
knutsenm, line 246: Oops, this should not be here

knutsenm (changes requested):
Also, rubocop has a couple tiny complaints

knutsenm (approved):

frett:
Don't merge this yet, I want to wait on merging this until we have a tangible use-case for it. Until then it's fine remaining approved and open.

db/schema.rb
```diff
@@ -258,22 +259,4 @@
...
-  add_foreign_key "follow_ups", "languages"
-  add_foreign_key "pages", "resources"
-  add_foreign_key "resources", "resource_types"
-  add_foreign_key "resources", "systems"
-  add_foreign_key "translated_attributes", "attributes"
-  add_foreign_key "translated_attributes", "translations"
-  add_foreign_key "translated_pages", "languages"
-  add_foreign_key "translated_pages", "resources"
-  add_foreign_key "translations", "languages"
-  add_foreign_key "translations", "resources"
```
frett, line 278: you dropped all the foreign keys when converting the values to an array field

frett:
@andrewroth @knutsenm I finally found a use case for this, so we can go ahead and fix the merge conflicts and get this merged into production.

knutsenm (changes requested):
Wrong review request?

Github says you have to resolve some merge conflicts here.

andrewroth:
Yeah, my bad. Wanted #1228

frett:
closing this since we never decided to move forward with it. If the need ever arises we can revive this PR/Branch

PR #847, opened 2022-01-07
[no-jira] Switch to codecov-action@v2

knutsenm:


.github/workflows/ruby.yml
```diff
@@ -4,7 +4,7 @@ on:
   pull_request:
     branches: [master]
   push:
-    branches: [staging]
+    branches: [master,staging]
```
knutsenm, line 7: This should restore coverage reports on merge to master

.github/workflows/ruby-cache.yml
```diff
@@ -1,17 +0,0 @@
...
-  cache:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Check out files
-        uses: actions/checkout@v2
-
-      - name: Set up Ruby
-        uses: ruby/setup-ruby@v1
-        with:
-          bundler-cache: true
```
knutsenm, line 17: No longer needed, since we cache in the `test` job that now runs against merge to master.

frett (approved):

PR #862, opened 2022-01-20
restrict valid characters for user counter names

frett:
This adds a validation to user counter names to reject any names that don't use URI safe characters

### DIFF
--- a/app/models/user_counter.rb
+++ b/app/models/user_counter.rb
@@ -4,7 +4,7 @@ class UserCounter < ApplicationRecord
   DECAY_RATE = -0.0077
 
   belongs_to :user
-  validates :counter_name, format: {with: /\A[a-zA-Z0-9_\-.]+\z/, message: "has invalid characters"}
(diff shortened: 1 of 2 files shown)


app/models/user_counter.rb
```diff
@@ -4,6 +4,7 @@ class UserCounter < ApplicationRecord
   DECAY_RATE = -0.0077
 
   belongs_to :user
+  validates :counter_name, format: {with: /\A[a-zA-Z0-9_\-.]+\z/, message: "has invalid characters"}
```
knutsenm, line 7: I wasn't sure about `\-` in a regex character class - I thought you were supposed to mention the hyphen first so it's not interpreted as a range character - but the docs I've read seem to indicate escaping it will work, too.

knutsenm (approved):

app/models/user_counter.rb
frett: I can move it to be the first character if that's a more standard practice in ruby/rails

PR #919, opened 2022-03-25
Upgrade to Rails 6.1

jbirdjavi:
This was a very simple upgrade (after coming from missionhub). All the changes are basically just default changes from Rails. Specs passed without any tweaks.


knutsenm (approved):
Looks utterly bog-standard ;-)

PR #935, opened 2022-04-14
GT-1434, more granular distribution

andrewroth:


config/routes.rb
```diff
@@ -42,6 +42,8 @@
 
   put "resources/:id/onesky", to: "resources#push_to_onesky"
 
+  get "/translations/files/:path", to: redirect("https://#{ENV.fetch("MOBILE_CONTENT_API_BUCKET")}.s3.#{ENV.fetch("AWS_REGION")}.amazonaws.com/%{path}", status: 302)
```
frett, line 45: for organization should we upload files to the s3 bucket under a `files` subdirectory or something similar?

spec/package_spec.rb
```diff
@@ -198,10 +201,10 @@
...
     let(:resources) do
       Nokogiri::XML('<resources xmlns="https://mobile-content-api.cru.org/xmlns/manifest">
-        <resource filename="mobile_only.png" src="2cf2ab68c49b217c6b2402699c742a236f96efe36bc48821eb6ba1a1427b8945"/>
-        <resource filename="web_mobile.png" src="4245551d69a8c582b6fc5185fb5312efc4f6863bda991a12a76102736f92fa2d"/>
-        <resource filename="both.png" src="ad03ee4cc7b015919b375539db150dee5f47245c6a293663c21c774b2dba294f"/>
-        <resource filename="wall.jpg" src="073d78ef4dc421f10d2db375414660d3983f506fabdaaff0887f6ee955aa3bdd"/>
+        <resource filename="mobile_only.png" src="2cf2ab68c49b217c6b2402699c742a236f96efe36bc48821eb6ba1a1427b8945.png"/>
+        <resource filename="web_mobile.png" src="4245551d69a8c582b6fc5185fb5312efc4f6863bda991a12a76102736f92fa2d.png"/>
+        <resource filename="both.png" src="ad03ee4cc7b015919b375539db150dee5f47245c6a293663c21c774b2dba294f.png"/>
+        <resource filename="wall.jpg" src="073d78ef4dc421f10d2db375414660d3983f506fabdaaff0887f6ee955aa3bdd.jpg"/>
```
frett, line 207: yay!! was looking to verify that the `src` attribute in the generated manifest file correctly included the file extensions now

config/routes.rb
frett: in that bucket we currently have:
- an attachments folder
- a GodTools folder (GodTools here references the "system" configured in the mobile-content-api database, there will be additional systems in the future)
- and a bunch of files in the root directory (I believe this is the new place attachments get written by rails)

I think we should put the files in a `/translations/files/` subdirectory within S3 to prevent us from continuting to overload the root directory with all different types of files

frett (changes requested):
can we put all the individual translation files that are uploaded to the S3 bucket into the `/translations/files/` subdirectory. that will help keep the S3 bucket organized because we upload the source attachments and the zips there as well

config/routes.rb
andrewroth: Sure, I can update the paths for that

andrewroth:
@frett (cc @knutsenm ) do you think `/translations/files/` should go into an env var or constant? If env var, something like `TRANSLATION_FILES_PATH=/translations/files/`?

frett:
I think a Constant should be good enough, I don't necessarily think we need it to be an environment variable that has to be configured for each environment

frett:
from my understanding of rails everything looks good to me. I'm going to try it out on the stage environment to make sure everything behaves correctly

app/lib/package.rb
```diff
@@ -14,11 +14,20 @@ class Package
   XPATH_TIPS = %w[//training:tip/@id
     //tract:header/@training:tip
     //tract:call-to-action/@training:tip].freeze
+  TRANSLATION_FILES_PATH = "/translations/files/"
```
frett, line 17:
```suggestion
  TRANSLATION_FILES_PATH = "translations/files/"
```

The leading `/` was generating a double-slash in S3. here is a sample file on stage:
```
https://cru-mobilecontentapi-staging.s3.amazonaws.com//translations/files/9c9715c5abb15fba40fde219fe8c89e7c5442235c397ee1e7bcc5a0fadc9c821.jpg
```

frett (approved):
From a functional standpoint everything looks good in the stage environment. @knutsenm any feedback on the rails code itself?

frett:
@knutsenm @andrewroth anything holding up this PR still?

knutsenm:
I haven't reviewed this yet. I'll do that now.

app/lib/package.rb
```diff
@@ -161,4 +170,12 @@ def include_tips?
     language_attributes = @translation.language.language_attributes.where(resource: @translation.resource)
     language_attributes.find_by(key: "include_tips")&.value == "true"
   end
+
+  def add_file_to_zip_and_s3(zip, zip_filename, local_filename)
```
knutsenm, line 174: Doesn't `sha_filename` remain a better name for the 2nd param?

knutsenm (changes requested):
Nothing serious, I'm just being my usual pedantic self ;-)

app/lib/package.rb
knutsenm: I mean the 2nd param, `zip_filename` - isn't it still a `sha_filename`?

andrewroth: Oh.. for that one, same idea, there's no enforcing that (which is the path of the file inside the zip archive) is actually going to be sha, so I was making the name to reflect that it could be anything

knutsenm: got it; sure!

spec/spec_helper.rb
```diff
@@ -107,15 +107,29 @@ def any_string
...
+      stub_const("ENV", "AWS_REGION" => region, "MOBILE_CONTENT_API_BUCKET" => bucket_name, "AWS_ACCESS_KEY_ID" => "aws_key", "AWS_SECRET_ACCESS_KEY" => "aws_secret")
 
       bucket = instance_double(Aws::S3::Bucket)
+
+      # record all s3 uploads to a hash so we can verify all files in the zip were sent to s3
+      @s3_uploads = {}
+      allow(bucket).to receive(:object) do |key|
+        new_object = Aws::S3::Object.new(key: key, bucket_name: bucket_name)
+        @s3_uploads[key] = {}
+        @last_object = new_object
```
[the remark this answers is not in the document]
andrewroth: Ah right.. was useful for my own debugging but I think we can delete it, and add it back in later if needed

knutsenm (approved):
What about using `Package::TRANSLATION_FILES_PATH` in more places?

andrewroth:
Oh, right..

knutsenm (approved):

config/routes.rb
```diff
@@ -42,7 +42,7 @@
 
   put "resources/:id/onesky", to: "resources#push_to_onesky"
 
-  get "/translations/files/:path",
+  get "/#{Package::TRANSLATION_FILES_PATH}:path",
```
frett, line 45: I would actually argue that `/translations/files/` here is distinct from the constant for S3.

spec/routes_spec.rb
```diff
@@ -2,14 +2,14 @@
 
 describe "routes", type: :request do
   it "redirects to s3" do
-    get "/translations/files/test.zip"
-    expect(response).to redirect_to("https://mybucket.s3.us-east-1.amazonaws.com/translations/files/test.zip")
+    get "/#{Package::TRANSLATION_FILES_PATH}test.zip"
```
frett, line 5: this one would also be distinct.

There are 2 distinct paths that happen to be the same, but don't necessarily need to be.
1. The path used for the mobile-content-api route
2. The path in S3 that you store the files in

frett (changes requested):
I'm overriding my previous approval to raise the point that we are using 1 constant for 2 distinct uses that just happen to be the same value currently.

frett:
> Oh, ok.. so basically just [8d71779](https://github.com/CruGlobal/mobile-content-api/commit/8d71779881da4be229a6c8ccdcaa3d082e93fb94) should be reversed

mostly, the lines that were updating the AWS urls can stay.
for example:
```
expect(response).to redirect_to("https://mybucket.s3.us-east-1.amazonaws.com/#{Package::TRANSLATION_FILES_PATH}test.zip")
```

frett (approved):

frett:
This should be good to merge

PR #980, opened 2022-05-31
Update rack (security update)

jbirdjavi:
Not sure why dependabot decided it couldn't open a PR for this.


frett (approved):

frett:
well, forcing dependabot to try again allowed it to correctly open a PR, no clue why it failed initially

frett:
I'm trying one more renovate setting that updates all transitive dependencies in the lockfile to see if that would have caught it.

PR #1054, opened 2022-08-25
GT-1776 move the user counters API to /user/me/counters

frett:
As we are starting to flesh out more user related APIs, I want to organize all user endpoints under: `/user/me` with a future potential url pattern of `/user/:user_id` to run the APIs against other users.

I kept a copy of the update route at the old location for old versions of the GodTools Android app

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -31,7 +31,11 @@ Rails.application.routes.draw do
   resources :custom_manifests, only: [:create, :update, :destroy, :show]
 
   scope "user" do
-    resources :counters, controller: "user_counters", only: [:index, :update]
+    resources :counters, controller: "user_counters", only: [:update] # Legacy route for GodTools Android v5.7.0-v6.0.0
+
+    scope "me" do
+      resources :counters, controller: "user_counters", only: [:index, :update]
+    end
   end
 
   get "monitors/lb"
--- a/spec/acceptance/user_counters_controller_spec.rb
+++ b/spec/acceptance/user_counters_controller_spec.rb
@@ -13,7 +13,7 @@ resource "UserCounters" do
 
   let(:structure) { FactoryBot.attributes_for(:user_counter)[:structure] }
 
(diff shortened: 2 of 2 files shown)


spec/acceptance/user_counters_controller_spec.rb
```diff
@@ -13,7 +13,7 @@
 
   let(:structure) { FactoryBot.attributes_for(:user_counter)[:structure] }
 
-  patch "user/counters/:id" do
+  patch "user/me/counters/:id" do
```
frett, line 16: Is there an easy way to run this test on both the legacy url and the new url?

frett: The options I can think of with my limited ruby/rails knowledge are:
- Duplicate the test code so there is a version for both endpoints
- loop over 2 paths (legacy & current) setting the path to a val and use it in the patch line

config/routes.rb
```diff
@@ -31,7 +31,11 @@
   resources :custom_manifests, only: [:create, :update, :destroy, :show]
 
   scope "user" do
-    resources :counters, controller: "user_counters", only: [:index, :update]
+    resources :counters, controller: "user_counters", only: [:update] # Legacy route for GodTools Android v5.7.0-v6.0.0
```
frett, line 34: I'm hoping we will only need to keep this legacy route for 6-12 months to give most users a chance to upgrade. Then we can delete this legacy route

spec/acceptance/user_counters_controller_spec.rb
knutsenm: I was just reading advice yesterday: test code need not be DRY, it should just be obvious. GA and copy/paste ;-)

frett:
This is behaving as expected in the stage environment

knutsenm (approved):

knutsenm (approved):

PR #1063, opened 2022-09-01
GT-1672, add abbreviation filter to resources

andrewroth:


frett (dismissed):
tests and lint are both failing right now

frett (commented):
everything appears to be functioning correctly in the stage environment. @knutsenm any comments on the rails logic itself?

knutsenm (approved):

PR #1073, opened 2022-09-16
GT-1562, track favoured tools

andrewroth:


app/controllers/favorite_tools_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def validate_ids
+    missing = tool_ids - Resource.pluck(:id).collect(&:to_s)
+    error_response = missing.collect{ |tool_id| [tool_id, "invalid tool id"] }
+    render(json: { errors: error_response }) if error_response.any?
+  end
+
+  def current_user_favorite_tools_response_json
+    { "data" => current_user.favorite_tools.collect { |fav| { "type" => "resource", "id" => fav.tool_id.to_s } } }
```
frett, line 38: does this not work using the standard jsonapi serializer?

knutsenm: Perhaps something about the requirements specified this JSON? (static "resource", no user_id)

knutsenm (changes requested):
Rubocop has some nits to pick

app/controllers/favorite_tools_controller.rb
frett: so, the requirements did include the sample jsonapi. the list of favorite tools is specifically a jsonapi resource identifier object pointing at a tool (`resource` within the mobile-content-api).

I was kind of expecting it to be seemless json generation like most of the rest of the API. But this is also the first time we've done a relationship only API endpoint like this

frett: @knutsenm in case you were curious, this is the relevant part of the jsonapi spec for this API: https://jsonapi.org/format/#crud-updating-to-many-relationships

frett: ah, I think it's not relying on the default serializer because it's using the `favorite_tools` relationship directly instead of resolving it to the actual `resource` objects.

app/controllers/favorite_tools_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    @tool_ids = params[:data].collect{ |data|
+      data[:id]
+    }
+  end
+
+  def validate_ids
+    missing = tool_ids - Resource.pluck(:id).collect(&:to_s)
+    error_response = missing.collect{ |tool_id| [tool_id, "invalid tool id"] }
+    render(json: { errors: error_response }) if error_response.any?
+  end
```
frett, line 35: something about this is brittle, if I send the `id` as a number it will fail, but if I send the same id as a string it will succeed.

I'd like this to not fail if the id is sent as a number.

frett: but if I use `"1"` it works correctly

app/serializers/user_serializer.rb
```diff
@@ -3,6 +3,7 @@
 class UserSerializer < ActiveModel::Serializer
   type "user"
   attributes :sso_guid, :created_at
+  has_many :favorite_tools, key: "favorite-tools"
```
frett, line 6: this is not being serialized correctly.

What I'm getting:
```
{
    "data": {
        "id": "1",
        "type": "user",
        "relationships": {
            "favorite-tools": {
                "data": [
                    {
                        "id": 1,
                        "tool-id": 1,
                        "user-id": 1,
                        "created-at": "2022-09-16T18:16:15.356Z",
                        "updated-at": "2022-09-16T18:16:15.356Z"
                    },
                    {
                        "id": 2,
                        "tool-id": 2,
                        "user-id": 1,
                        "created-at": "2022-09-16T18:16:15.379Z",
                        "updated-at": "2022-09-16T18:16:15.379Z"
                    },
                    {
                        "id": 3,
                        "tool-id": 10,
                        "user-id": 1,
                        "created-at": "2022-09-16T18:24:57.745Z",
                        "updated-at": "2022-09-16T18:24:57.745Z"
                    }
                ]
            }
        }
    }
}
```

What I'm expecting:
```
{
    "data": {
        "id": "1",
        "type": "user",
        "relationships": {
            "favorite-tools": {
                "data": [
                    {
                        "type": "resource",
                        "id": 1
                    },
                    {
                        "type": "resource",
                        "id": 2
                    },
                    {
                        "type": "resource",
                        "id": 10
                    }
                ]
            }
        }
    }
}
```

frett: take note that the `id` being returned should be the id of the tool itself, and not the favorite-tool relationship

frett: this is on a `GET /users/me` request

frett (changes requested):
there are several correctness issues with the actual API requests and responses. See the comments about several things not currently working correctly

frett:
I'm getting a 404 error on `DELETE /users/me/relationships/favorite-tools` request, even when I account for the currently incorrect route.

app/controllers/favorite_tools_controller.rb
frett: I'm fine with this logic persisting as is because it appears to allow a DB optimization.

app/serializers/user_serializer.rb
frett: this is supposed to represent a many-to-many relationship. Currently you are serializing the intermediate internal db object tracking that many-to-many relationship and not the actual objects on the other side of the relationship.

app/controllers/favorite_tools_controller.rb
andrewroth: Oh, right, I'm used to the regular rails params that has all strings. These params are automatically parsed from the json so they can be ints.

andrewroth:
Ah, right, the default rails delete routes require an ID param, so I'll add a custom route

knutsenm (approved):

app/controllers/favorite_tools_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    @tool_ids = params[:data].collect { |data|
+      data[:id].to_s
+    }
+  end
+
+  def validate_ids
+    missing = tool_ids - Resource.pluck(:id).collect(&:to_s)
+    error_response = missing.collect { |tool_id| [tool_id, "invalid tool id"] }
+    render(json: {errors: error_response}) if error_response.any?
```
frett, line 35: I know this was an initial placeholder for the error response. Yesterday I updated the jira ticket to contain the desired error response when an invalid id is specified. here is what the response json should look like:

```json
{
  "errors" : [
    {
      "code": "invalid_tool",
      "meta": {
        "tool_id": 100
      }
    },
    {
      "code": "invalid_tool",
      "meta": {
        "tool_id": 30
      }
    }
  ]
}
```

frett: this should also return a 400 Bad Request response instead of the 200 OK response it's currently returning

frett (approved):
I believe everything looks good. Feel free to merge this, I'll open new jira tickets if anything new shows up

frett (approved):

PR #1090, opened 2022-10-05
GT-1812 expose the user-id when authenticating a user

frett:


app/models/auth_token.rb
```diff
@@ -64,4 +64,12 @@ def token
...
     24.hours.from_now
   end
+
+  def user_id
+    if @user
+      @user.id
+    else
+      nil
+    end
+  end
```
frett, line 74: is there a cleaner way to do this?

frett:
This is working correctly in the stage environment

knutsenm (approved):

PR #1119, opened 2022-10-29
GT-1801, implement user delete api

andrewroth:


app/controllers/users_controller.rb
```diff
@@ -1,9 +1,19 @@
...
-
   def show
     render json: @user
   end
+
+  def destroy
+    @user.user_counters.destroy_all
+    @user.favorite_tools.destroy_all
+    render json: "", status: 204
+  end
```
frett, line 12: does this also delete the user object?

andrewroth: Oops! Forgot the most important thing. I'll add that in.

frett:
adding @knutsenm for the rails review. I'm in the process of testing this out on the stage environment

app/controllers/application_controller.rb
```diff
@@ -68,7 +68,13 @@ def current_user
   def render_unauthorized
     authorization = AccessCode.new
     authorization.errors.add(:id, "Unauthorized")
-    render_error(authorization, :unauthorized)
+    render_error(authorization, :unauthorized) # 401
+  end
+
+  def render_forbidden
+    authorization = AccessCode.new
+    authorization.errors.add(:id, "Frobidden")
```
frett, line 76:
```suggestion
    authorization.errors.add(:id, "Forbidden")
```

frett:
Functionally deleting a user works, and I've been unable to trick it into deleting a different user. But it appears that error responses aren't always accurate.

Scenario 1:
1. Have a valid auth token
2. Try getting information about a user that doesn't exist (e.g. `GET /users/999999`) returns 404 instead of the expected 403 forbidden

Scenario 2:
1. Have a valid auth token
2. Delete your user account (leaves auth token correctly signed, but pointing at an invalid user)
3. Trying to delete another user by id instead of using `me` yields either a 403 or 404 response, it should be a 401 response because the auth token is no longer valid.

frett (approved):
Everything checks out for me, going to give @knutsenm a chance to review the ruby/rails before we merge this

app/controllers/with_user_controller.rb
```diff
@@ -4,16 +4,29 @@ class WithUserController < ApplicationController
...
 
-    # currently, if trying to operate on a specific @user, you can only operate on your own user data, but this may change later
+    # can always operate on your own user
     return if @user && @user == current_user
 
+    # currently, if trying to operate on a specific @user, you can only operate on your own user data, but this may change later
+    render_forbidden and return if @user.nil? || @user != current_user
+
     render_unauthorized
   end
```
frett, line 24: I believe `authorize` can be simplified to this:
```ruby
  def authorize!
    # make sure we have a valid auth to begin with
    render_unauthorized and return unless current_user_id && current_user && authorization

    # all with user controllers also operate on a subject, sometimes "me".
    # we want this here, after the token is checked, but before the subject authorization is checked
    @user = params[user_id_attribute].blank? || params[user_id_attribute] == "me" ? current_user : User.find_by(id: params[user_id_attribute])

    # currently, if trying to operate on a specific @user, you can only operate on your own user data, but this may change later
    render_forbidden and return if @user != current_user
  end
```

frett: When I change it out locally all the current tests pass and it looks logically correct

app/controllers/users_controller.rb
```diff
@@ -1,9 +1,20 @@
...
-  prepend_before_action { set_user(:id) }
-
   def show
     render json: @user
   end
+
+  def destroy
+    @user.user_counters.destroy_all
+    @user.favorite_tools.destroy_all
+    @user.destroy
```
knutsenm, line 11: Don't you usually do `.destroy!` so as to throw errors? Most of the other controllers do.

spec/acceptance/users_controller_spec.rb
```diff
@@ -84,4 +90,59 @@
...
+        do_request
+        expect(status).to eq(403)
+      end
+    end
+    context "self deleted" do
+      requires_okta_login
+
+      it "returns 401" do
+        user.destroy
+        do_request
```
knutsenm, line 143: `:id` is not set for this test

knutsenm (changes requested):
Minor nits

app/controllers/with_user_controller.rb
```diff
@@ -4,16 +4,21 @@ class WithUserController < ApplicationController
   before_action :authorize!
 
   def authorize!
-    # requested is authorized if using okta to provide a valid user id, and no specific user is set
-    return if authorization && current_user && !@user
+    # make sure we have a valid auth to begin with
+    render_unauthorized and return unless authorization && current_user_id && current_user
```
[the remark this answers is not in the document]
andrewroth: This seems more confusing than what's there now to me... the flow of the method is return (valid) if ... and I think it's easier to understand that way

app/controllers/with_user_controller.rb
```diff
@@ -4,16 +4,21 @@ class WithUserController < ApplicationController
...
 
-    # currently, if trying to operate on a specific @user, you can only operate on your own user data, but this may change later
-    return if @user && @user == current_user
+    # all with user controllers also operate on a subject, sometimes "me".
+    # we want this here, after the token is checked, but before the subject authorization is checked
+    @user = params[user_id_attribute].blank? || params[user_id_attribute] == "me" ? current_user : User.find_by(id: params[user_id_attribute])
```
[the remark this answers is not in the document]
andrewroth: @knutsenm standardrb directly contradicts you:

== app/controllers/with_user_controller.rb ==
C: 12: 13: [Correctable] Style/TernaryParentheses: Omit parentheses for ternary conditions.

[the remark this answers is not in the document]
knutsenm: Well, I wouldn't want to contradict rubocop, would I ;-)

I'll just have to live with my second guessing.

knutsenm (approved):
Are you going to edit FavoriteToolsController? Your choice.

> should FavoriteToolsController be using `@user` rather than `current_user`, like the other WithUserController(s) do?

andrewroth:
I would say let's leave it. Can be added in later

frett:
This is good to merge whenever

app/controllers/with_user_controller.rb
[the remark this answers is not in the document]
knutsenm: As of today, `standardrb` agrees w/me ;-)

knutsenm (approved):

PR #1122, opened 2022-11-03
Update for ECS multi-container

Omicron7:


frett:
The current CORS configuration is preventing uploading new attachments from working. I would assume this is related to the rack-cors configuration.

For reference the following permissions would need to be covered:
- mobilecontentadmin(-stage).cru.org needs credentials access in the CORS configuration
- All other hosts don't need credentials access, but they do need generic CORS access to lookup which tools exist and access tool content

The error I'm seeing when trying to uploading a new attachment is 
Screen Shot 2022-11-04 at 12 04 42 PM

jbirdjavi (approved):

PR #1129, opened 2022-11-08
GT-1797, translated/translation attributes

andrewroth:


andrewroth:
@frett @knutsenm not done yet, but open for initial comments

app/models/translation.rb
```diff
@@ -95,18 +95,36 @@ def prevent_destroy_published
   def push_published_to_s3
     return unless is_published
 
-    name_desc_onesky if resource.uses_onesky?
+    phrases = manifest_translated_phrases
```
frett, line 98: this should probably be protected by `resources.uses_onesky?`

app/models/translation.rb
```diff
@@ -95,18 +95,36 @@ def prevent_destroy_published
...
+      if translation.present?
+        translation_attribute = translation_attributes.where(key: translated_attributes.key).first_or_create
+        translation_attribute.update(value: translation)
+        translation_attribute_ids << translation_attribute.id
+      elsif translated_attribute.required
+        raise Error::TextNotFoundError, "Translated phrase not found: ID: #{translated_attribute.onesky_phrase_id}"
+      end
+    end
+
+    translation_attributes.where.not(id: translation_attribute_ids).delete_all
```
frett, line 127: I'm not sure if we need to delete other `translation_attributes` not added by this process. For now I'm indifferent to this, but this could have future impact if we ever add an additional mechanism to populate other `translation_attributes`

app/models/translation.rb
```diff
@@ -95,18 +95,36 @@ def prevent_destroy_published
...
+    translation_attribute_ids = []
+    resource.translated_attributes.each do |translated_attribute|
+      translation = phrases[translated_attribute.onesky_phrase_id]
+
+      if translation.present?
+        translation_attribute = translation_attributes.where(key: translated_attributes.key).first_or_create
+        translation_attribute.update(value: translation)
+        translation_attribute_ids << translation_attribute.id
+      elsif translated_attribute.required
+        raise Error::TextNotFoundError, "Translated phrase not found: ID: #{translated_attribute.onesky_phrase_id}"
```
frett, line 123: I'm guessing this doesn't trigger a rollback of other attributes when a missing attribute is encountered? or is this all in a db transaction so a rollback happens automatically?

andrewroth: Good question, I believe the attributes already created would remain. I can manually enclose it in a transaction, though. Should I?

frett: I would put any updates to the translation object in a single transaction so it rolls back if there is a failure somewhere.

This would include storing the name and description on the translation object, and even storing the manifest file name

app/models/translation.rb
andrewroth: Right, good call

app/models/translation.rb
andrewroth: Up to you. I just figure since the API will return a list of all of them, it's most accurate to include only the ones from the latest translation. Easy to remove if you prefer.

frett: We only publish a specific translation object once, the next time we publish that tool in that language it's on a new translation object.

andrewroth: Ok, in that case, it's not needed

andrewroth:
Ok, I've added tests. From my perspective I think this is ready, pending code review tweaks.

frett:
I'm getting a 404 error issuing the follow request:

```
POST /resources/3/translated-attributes HTTP/1.1
Host: mobile-content-api-stage.cru.org
Accept: application/vnd.api+json
Content-Length: 206

{
    "data": {
        "type": "translated-attribute",
        "attributes": {
            "name": "outline",
            "onesky-phrase-id": "info.outline",
            "required": false
        }
    }
}
```

app/models/translation_attribute.rb
```diff
@@ -0,0 +1,3 @@
+class TranslationAttribute < ApplicationRecord
+  belongs_to :translation
```
knutsenm, line 2: Does this need the validations (on `value` and `translation`) that the model used to have pre-rename?

db/schema.rb
```diff
@@ -252,7 +257,9 @@
     t.date "last_decay", default: -> { "now()" }
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
+    t.string "values", default: [], array: true
     t.index ["user_id", "counter_name"], name: "index_user_counters_on_user_id_and_counter_name", unique: true
+    t.index ["values"], name: "index_user_counters_on_values", using: :gin
```
knutsenm, line 262: Where did `values` come from?

db/schema.rb
```diff
@@ -151,8 +151,6 @@
     t.string "key", null: false
     t.string "value", null: false
     t.boolean "is_translatable", default: false
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
```
knutsenm, line 155: I don't see a migration removing those

knutsenm (changes requested):

frett:
I manually added a couple translated_attributes to the database and was able to successfully test the publishing process and the enforcement of the `required` flag and raising appropriate errors.

frett:
The routes to manage translated_attributes aren't currently configured at the correct endpoints. They should all be under the resource object they are scoped to.

```
POST /resources/1/translated-attributes
PUT /resources/1/translated-attributes/2
DELETE /resources/1/translated-attributes/3
```

app/models/translation.rb
```diff
@@ -94,18 +95,40 @@ def prevent_destroy_published
   def push_published_to_s3
     return unless is_published
 
-    name_desc_onesky if resource.uses_onesky?
+    ActiveRecord::Base.transaction do
+      if resource.uses_onesky?
+        phrases = manifest_translated_phrases
+        name_desc_onesky(phrases)
+        create_translated_attributes(phrases)
```
frett, line 102: hmm, these could still leak partial attributes if `Package.new(self).push_to_s3` fails. But I'm not sure of a great solution to this because the `push_to_s3` process shouldn't be part of the transaction.

For now this is fine unless you have a potential quick solution in mind

frett:
There currently isn't consistency in the `onesky-phrase-id` attribute name. when reading the value from json the name is: `onesky-phrase-id`, when creating a new `translated-attribute` or updating an existing one the attribute name needs to be `onesky_phrase_id`.

These attribute names should be consistent between reads and writes. I would prefer the attribute name be `onesky-phrase-id` for API interactions with `translated-attribute` objects to follow the `kebab-naming` convention.

I do know we have similar inconsistencies in other APIs that exist, at some point we should probably clean those up, but that's out of the scope of this ticket and PR.

frett (changes requested):
we need a consistent attribute name for `onesky-phrase-id`

app/models/translation_attribute.rb
andrewroth: No need for translation, rails now has belongs_to as required by default, I believe. Adding one on value isn't a bad idea, though.

frett:
`key` on the `translated-attribute` model needs to be unique within the scope of a resource

frett:
`key` should also have a non-null validation on it, when I was testing out the create endpoint I was missing `key` and it successfully created an attribute with the `key` equaling `null`

db/schema.rb
andrewroth: Oops, I'll put those back

frett:
on creation of a new `translated-attribute`, `required` should default to `false`, it current defaults to `null`

frett:
we should probably also make `onesky-phrase-id` non-null

app/controllers/translated_attributes_controller.rb
```diff
@@ -18,7 +18,19 @@ def destroy
...
   def create_translated_attr
-    a = @resource.translated_attributes.create!(permitted_params)
+    a = @resource.translated_attributes.create(permitted_params)
+
+    if a.errors[:key] == ["has already been taken"]
+      render(json: {errors: { "code" => "key_already_exists"}}, status: 400)
+    elsif a.errors[:key] == ["can't be blank"]
+      render(json: {errors: { "code" => "invalid_key"}}, status: 400)
+    elsif a.errors[:onesky_phrase_id] == ["can't be blank"]
+      render(json: {errors: { "code" => "invalid_onesky_phrase_id"}}, status: 400)
```
frett, line 28: these errors should also be returned for updating a translated_attr

app/controllers/translated_attributes_controller.rb
```diff
@@ -16,25 +18,48 @@ def destroy
...
     head :no_content
   end
 
+  def render_translation_attribute_error(a)
+    if a.errors[:key] == ["has already been taken"]
+      render(json: {errors: {"code" => "key_already_exists"}}, status: 400)
+    elsif a.errors[:key] == ["can't be blank"]
+      render(json: {errors: {"code" => "invalid_key"}}, status: 400)
+    elsif a.errors[:onesky_phrase_id] == ["can't be blank"]
+      render(json: {errors: {"code" => "invalid_onesky_phrase_id"}}, status: 400)
```
frett, line 43: There currently isn't consistency in the `onesky-phrase-id` attribute name. when reading the value from json the name is: `onesky-phrase-id`, when creating a new `translated-attribute` or updating an existing one the attribute name needs to be `onesky_phrase_id`.

These attribute names should be consistent between reads and writes. I would prefer the attribute name be `onesky-phrase-id` for API interactions with `translated-attribute` objects to follow the `kebab-naming` convention.

I do know we have similar inconsistencies in other APIs that exist, at some point we should probably clean those up, but that's out of the scope of this ticket and PR.

frett: when submitting this create request:

```
POST /resources/3/translated-attributes HTTP/1.1
Host: mobile-content-api-stage.cru.org
Accept: application/vnd.api+json
Content-Length: 175

{
    "data": {
        "type": "translated-attribute",
        "attributes": {
            "key": "outline2",
            "onesky-phrase-id": "info.outline"
        }
    }
}
```

I get back this error:
```
{
    "errors": {
        "code": "invalid_onesky_phrase_id"
    }
}
```

frett (approved):
From a functional standpoint everything looks good

knutsenm (approved):

frett:
I'm good with merging this whenever

frett:
@andrewroth @knutsenm is there anything holding up merging this PR?

knutsenm:
> @andrewroth @knutsenm is there anything holding up merging this PR?

No; go for it!

PR #1217, opened 2023-02-21
GT-1939 add the appropriate key to the `default_variant` serializer relationship

frett:
without this, using `include=default-variant` doesn't include the relationship


frett:
I've tested this behavior before & after on the stage environment

knutsenm (approved):

PR #1228, opened 2023-03-08
GT-1839, implement facebook auth

andrewroth:


app/controllers/auth_controller.rb
```diff
@@ -30,4 +36,32 @@ def auth_with_okta
...
+      data = JSON.parse(Net::HTTP.get(URI(url)))
+    rescue Net::HTTPBadRequest => e
+      render_bad_request(e.to_s) and return
+    end
+
+    (render_bad_request(data["data"]["error"]) and return) if data["data"] && data["data"]["error"]
+    (render_bad_request("facebook token is not valid") and return) unless data["data"] && data["data"]["is_valid"] && data["data"]["user_id"]
+
+    user_id = data["data"]["user_id"]
+    url = "https://graph.facebook.com/#{user_id}?fields=email,id,first_name,last_name,short_name&access_token=#{data_attrs[:facebook_access_token]}"
```
frett, line 54: you will need to include the secret here, dev currently has the "require app secret" setting disabled, but we will want that enabled, and it is currently enabled on production.

app/controllers/auth_controller.rb
```diff
@@ -30,4 +36,32 @@ def auth_with_okta
...
+    begin
+      data = JSON.parse(Net::HTTP.get(URI(url)))
+    rescue Net::HTTPBadRequest
+      render_bad_request("could not extract facebook fields required") and return
+    end
+
+    (render_bad_request(data["data"]["error"]) and return) if data["data"] && data["data"]["error"]
+
+    user = User.where(email: data["email"]).first_or_create
+    user.update!(facebook_user_id: data["id"], first_name: data["first_name"], last_name: data["last_name"], short_name: data["short_name"])
```
frett, line 64: I would like to default the user lookup to prefer looking up the user by the `facebook_user_id` first. I think I'm fine with a fallback to looking the user up by email if the user_id doesn't match.

frett: side note, is `email` listed as needing to be unique in the `User` model? we might need to relax that if so because I could come up with a scenario where 2 accounts would have the same email address based on a user updating email addresses/creating a new account.

frett: falling back to email address lookup could have other implications for how data is linked to individual accounts. so just do `facebook_user_id` matching for now.

frett: Scenario 1:

1. `fbUser1` logs in with `email1`
   - This create `mcUser1`
2. `fbUser1` changes the email on their facebook account to `email2`
3. `fbUser2` creates/updates their account to use `email1` as their email
4. `fbUser2` logs in with `email1`
   - The current algorithm will find `mcUser1` for `fbUser2` and return that as the logged in user

app/controllers/auth_controller.rb
andrewroth: Surprisingly, it works in production without app secret. I just tested it.

app/controllers/auth_controller.rb
andrewroth: Ah, right, ok. Makes sense.

app/controllers/auth_controller.rb
frett: ok, guess we don't need to do anything different here then

frett:
I was able to use this to successfully log in to the stage environment

app/services/facebook.rb
```diff
@@ -0,0 +1,63 @@
...
+      user.update(user_info)
+      user
+    end
+
+    def create_user(user_info)
+      User.create!(user_info)
+    end
+
+    def validate_and_extract_user_id(access_token)
+      unless /^[0-9a-zA-Z]+$/.match?(access_token)
```
frett, line 29: just curious, did they specify somewhere in their documentation that facebook access_tokens follow this format?

andrewroth: I couldn't find anything on that, just from observation of your token. I guess that's a pretty small sample heh. I'm just concerned about the security hole of someone adding & or ? in their token and then they're able to pass other params.

frett: rails doesn't provide any sort of uri escaping functionality to allow you to set url parameters safely no matter what characters they contain?

frett: it looks like in order to get escaping you should be sending the query parameters as a second parameter.

https://github.com/jnunemaker/httparty#examples

frett: @andrewroth bumping my response :)

frett:
it looks like the deletion_request API works correctly for facebook. I was able to initiate the delete from my facebook account and verify the account was deleted from the API.

db/migrate/20230310151243_create_deletion_request.rb
```diff
@@ -0,0 +1,12 @@
+class CreateDeletionRequest < ActiveRecord::Migration[6.1]
+  def change
+    create_table :deletion_requests do |t|
+      t.string :provider
+      t.string :uid
+      t.string :pid
```
knutsenm, line 6: What does that stand for, "provider ID"?

app/services/facebook.rb
```diff
@@ -0,0 +1,59 @@
...
+    def validate_and_extract_user_id(access_token)
+      url = "https://graph.facebook.com/debug_token"
+      data = JSON.parse(get(url, query: {input_token: access_token, access_token: "#{ENV.fetch("FACEBOOK_APP_ID")}|#{ENV.fetch("FACEBOOK_APP_SECRET")}"}))
+      raise Facebook::FailedAuthentication, "Error validating access_token with Facebook: #{data["data"]["error"]}" if data["data"] && data["data"]["error"]
+      raise Facebook::FailedAuthentication, "Error validating access_token with Facebook: no facebook user id returned" unless data["data"] && data["data"]["is_valid"] && data["data"]["user_id"]
+
+      data["data"]["user_id"]
+    end
+
+    def transform_fields(fields)
```
knutsenm, line 37: This, too, is only called once...

app/models/deletion_request.rb
```diff
@@ -0,0 +1,57 @@
...
+    exp = OpenSSL::HMAC.digest("SHA256", ENV["FACEBOOK_APP_SECRET"], payload)
+    raise FailedAuthentication, "FB deletion callback called with invalid data" if decoded != exp
+
+    data
+  end
+
+  private
+
+  def associated_user
+    # more providers will be added
```
knutsenm, line 38: How many more? That will affect how this logic grows, & maybe how many more `User#*_user_id` attributes you add.

app/services/facebook.rb
```diff
@@ -0,0 +1,59 @@
...
+        facebook_user_id: fields["id"],
+        email: fields["email"],
+        first_name: fields["first_name"],
+        last_name: fields["last_name"],
+        short_name: fields["short_name"]
+      }.with_indifferent_access
+    end
+
+    def fetch_account_profile(access_token, user_id)
+      url = "https://graph.facebook.com/#{user_id}"
```
knutsenm, line 48: Maybe `https://graph.facebook.com` should be a constant, or a secret?

knutsenm (changes requested):
This looks really good overall. I only have minor comments and questions.

app/services/facebook.rb
frett: Secret doesn't make sense because I don't see Facebook ever changing their public API domain.

In the Okta login service it uses a `base_uri` setting that could extract the root API path to avoid the duplication between the 2 API requests

db/migrate/20230310151243_create_deletion_request.rb
andrewroth: public id

andrewroth: it's the code that can be used to query to check the status

app/controllers/auth_controller.rb
```diff
@@ -6,6 +6,8 @@ def create
       auth_with_okta
     elsif data_attrs[:facebook_access_token]
       auth_with_facebook
+    elsif data_attrs[:google_access_token]
```
frett, line 9: are you using an `access_token` or an `id_token` for google authentication?

frett:
I would have preferred for us to finish up the facebook part of this PR and merge that without needing to bundle the google auth stuff with it

app/controllers/auth_controller.rb
frett: because we are using an `id_token` to authenticate for google, this attribute should be `google_id_token`

app/controllers/auth_controller.rb
```diff
@@ -1,9 +1,35 @@
...
   def create
-    token = data_attrs[:okta_access_token] ? auth_with_okta : auth_with_code
+    method = THIRD_PARTY_AUTH_METHODS.keys.detect { |method| data_attrs[:"#{method}_access_token"] }
+    token = case method
+    when :apple
+      # special case for apple, which has given and family name passed in
+      user = Apple.find_user_by_access_token(data_attrs[:apple_access_token], data_attrs[:apple_given_name], data_attrs[:apple_family_name])
+      AuthToken.new(user: user)
+    when :okta, :facebook, :google
+      user = THIRD_PARTY_AUTH_METHODS[method].find_user_by_access_token(data_attrs[:"#{method}_access_token"])
```
frett, line 19: for google, we are passing in an `id_token`, so the data attribute should be `google_id_token`

app/controllers/auth_controller.rb
```diff
@@ -1,9 +1,35 @@
...
+    apple: Apple
+  }
+
   def create
-    token = data_attrs[:okta_access_token] ? auth_with_okta : auth_with_code
+    method = THIRD_PARTY_AUTH_METHODS.keys.detect { |method| data_attrs[:"#{method}_access_token"] }
+    token = case method
+    when :apple
+      # special case for apple, which has given and family name passed in
+      user = Apple.find_user_by_access_token(data_attrs[:apple_access_token], data_attrs[:apple_given_name], data_attrs[:apple_family_name])
```
frett, line 16: according to @levieggertcru , we are passing an `id_token` for apple auth, so the parameter should be named `apple_id_token` instead of `apple_access_token`

app/controllers/auth_controller.rb
```diff
@@ -1,9 +1,35 @@
...
+      AuthToken.new(user: user)
+    when :okta, :facebook, :google
+      user = THIRD_PARTY_AUTH_METHODS[method].find_user_by_access_token(data_attrs[:"#{method}_access_token"])
+      AuthToken.new(user: user)
+    else
+      AccessCode.validate(data_attrs[:code])
+      AuthToken.new
```
frett, line 23: No real action point here, but just letting you know that the `AccessCode` logic has been slated to be removed from mobile-content-api since November 2021. See GT-1348

app/models/deletion_request.rb
knutsenm: OK, is this a TBD?

app/services/facebook.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+class Facebook < AuthServiceBase
```
knutsenm, line 3: This class could be named `FacebookAuth`. A shame, however, that `AppleAuth` is already taken.

app/services/okta.rb
```diff
@@ -1,64 +1,64 @@
 # frozen_string_literal: true
 
-class Okta
-  include HTTParty
+class Okta < AuthServiceBase
```
knutsenm, line 3: This could be named `OktaAuth`

db/schema.rb
```diff
@@ -270,6 +279,12 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.boolean "admin", default: false, null: false
+    t.string "short_name"
+    t.string "facebook_user_id"
+    t.string "gr_master_person_id"
+    t.string "google_user_id"
+    t.string "apple_user_id"
```
knutsenm, line 286: Now that there are three *_user_id, might there eventually be four? Would it be better to have a single user_id field and a type field?

app/models/user.rb
```diff
@@ -1,9 +1,9 @@
 class User < ApplicationRecord
-  has_many :user_counters
-  has_many :favorite_tools
+  has_many :user_counters, dependent: :destroy
+  has_many :favorite_tools, dependent: :destroy
   has_many :tools, through: :favorite_tools
 
-  validates :sso_guid, uniqueness: true, presence: true
+  validates :sso_guid, uniqueness: true, presence: true, unless: -> { true }
```
knutsenm, line 6: Wha?
```
unless: -> { true }
```

app/services/google_auth.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+class GoogleAuth < AuthServiceBase
+  include HTTParty
```
knutsenm, line 4: No longer needed here.

db/schema.rb
frett: I'd like to reserve the possibility to have a user linked to multiple identity providers in the future. Consolidating this into a `user_id` and `type` field would prevent that from happening without future refactoring

knutsenm: 👍

app/models/deletion_request.rb
andrewroth: I'm not sure, but I would be surprised if there's more after this. @frett do you know? Can probably remove this comment now

app/services/facebook.rb
andrewroth: We could just add Service to the end. So `AppleAuthService` and `FacebookAuthService`...

app/models/user.rb
andrewroth: Ooops, this was a test I was doing and didn't mean to commit that

app/services/google_auth.rb
```diff
@@ -16,27 +16,31 @@ def service_name
...
 
     def remote_user_id(decoded_token)
       decoded_token["sub"]
     end
 
-    def extract_user_atts(access_token, decoded_token)
+    def extract_user_atts(_google_id_token, decoded_token)
       {
-        google_user_id: remote_user_id(decoded_token),
+        google_id_token: remote_user_id(decoded_token),
```
frett: this should have stayed as `google_user_id`

app/models/user.rb
```diff
@@ -3,7 +3,7 @@ class User < ApplicationRecord
   has_many :favorite_tools, dependent: :destroy
   has_many :tools, through: :favorite_tools
 
-  validates :sso_guid, uniqueness: true, presence: true, unless: -> { facebook_user_id.present? || google_user_id.present? || apple_user_id.present? }
+  validates :sso_guid, uniqueness: true, presence: true, unless: -> { facebook_user_id.present? || google_id_token.present? || apple_id_token.present? }
```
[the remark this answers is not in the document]
andrewroth: Right, I realized that earlier and should be fixed now

app/services/google_auth.rb
andrewroth: Same for this

app/services/google_auth.rb
andrewroth: Fixed

frett:
The latest version of this PR is failing to deploy to the stage environment, I think it has to do with some of the changing DB migrations. It might take manually fixing the stage database to get this to deploy

app/services/apple_auth_service.rb
```diff
@@ -0,0 +1,52 @@
...
+      %w[sub email iss aud]
+    end
+
+    def decode_token(apple_id_token)
+      AppleAuth::JWTDecoder.new(apple_id_token).call
+    end
+
+    def validate_token!(apple_id_token, decoded_token)
+      raise FailedAuthentication, "Sub is missing from payload" unless decoded_token["sub"]
+      AppleAuth::UserIdentity.new(decoded_token["sub"], apple_id_token).validate!
```
frett, line 35: @andrewroth I believe we actually want to validate the `aud` field here, and not the `sub`.

@levieggertcru any input?

app/services/facebook_auth_service.rb
```diff
@@ -0,0 +1,59 @@
...
+    def expected_fields
+      %w[user_id]
+    end
+
+    def expected_fields_from_get_fields
+      %w[id email]
+    end
+
+    def decode_token(input_token)
+      decoded_token = JSON.parse(get("/debug_token", query: {input_token: input_token, access_token: "#{ENV.fetch("FACEBOOK_APP_ID")}|#{ENV.fetch("FACEBOOK_APP_SECRET")}"}))
```
frett, line 19: https://app.rollbar.com/a/Cru/fix/item/mobile-content-api/285

frett (changes requested):
Facebook auth isn't working currently, getting a rollbar error when validating an access_token

knutsenm (changes requested):

db/schema.rb
```diff
@@ -259,7 +268,9 @@
     t.date "last_decay", default: -> { "now()" }
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
+    t.string "values", default: [], array: true
     t.index ["user_id", "counter_name"], name: "index_user_counters_on_user_id_and_counter_name", unique: true
+    t.index ["values"], name: "index_user_counters_on_values", using: :gin
```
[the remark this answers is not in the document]
frett: good catch Mark, it's from this PR: https://github.com/CruGlobal/mobile-content-api/pull/826

which we've put on indefinite hold because we have no actual use cases that require it

frett:
@levieggertcru confirmed that the Apple login functionality is working on stage

db/schema.rb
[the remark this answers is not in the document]
knutsenm: Andrew, that must have snuck into your local db; you should revert it.

frett:
Current error when logging in with Facebook:

```
{"errors":[{"source":{"pointer":"/data/attributes/id"},"detail":"Validation failed: Email has already been taken"}]}
```

We need to remove the unique constraint on user.email because accounts from different login providers could share the same email address.

Also merging users based on email address is problemsome due to: https://github.com/CruGlobal/mobile-content-api/pull/1228#discussion_r1129909467

frett:
facebook login is working again. so that should be all 3 of them tested.

From a functional perspective this PR looks good

frett (approved):

frett:
I'm hoping to get this merged by the end of the week

app/services/auth_service_base.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+class AuthServiceBase
+  include HTTParty
+
+  class << self
+    def find_user_by_token(access_token)
+      decoded_token = decode_token(access_token)
```
knutsenm, line 8: It would be nice if must-implement functions like `decode_token` and `validate_token!` and `extract_user_atts` were stubbed out here (& throw an error)

app/services/auth_service_base.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+class AuthServiceBase
```
knutsenm, line 3: `BaseAuthService`, perhaps?

app/services/google_auth_service.rb
```diff
@@ -0,0 +1,48 @@
...
+      super
+    rescue Google::Auth::IDTokens::ExpiredTokenError => e
+      raise self::FailedAuthentication, e.message
+    end
+
+    private
+
+    def service_name
+      "google"
+    end
```
knutsenm, line 15: Is this necessary? Looks like the inherited method might work.

app/services/okta_auth_service.rb
```diff
@@ -0,0 +1,64 @@
...
+      unless decoded_token["iss"] == ENV["OKTA_SERVER_PATH"]
+        raise self::FailedAuthentication, "Invalid issuer. Expected #{ENV["OKTA_SERVER_PATH"]}, received #{decoded_token["iss"]}"
+      end
+    end
+
+    def remote_user_id(decoded_token)
+      # noop since remote_user_id for okta is ssoguid given in the userinfo call
+    end
+
+    # manually implement setup_user to pull ssoguid from usre_atts instead of using remote_user_id
```
knutsenm, line 38: I think what you mean is

```suggestion
    # reimplement setup_user to search on sso_guid (from user_atts) instead of using remote_user_id
```

app/services/okta_auth_service.rb
```diff
@@ -0,0 +1,64 @@
...
+    def setup_user(remote_user_id, user_atts)
+      user = User.where(sso_guid: user_atts[primary_key]).first_or_initialize
```
knutsenm, line 40: Either go all out,

```suggestion
      user = User.where("#{primary_key}": user_atts[primary_key]).first_or_initialize
```

(but I'm not sure of that syntax), or make `def primary_key` empty and

```suggestion
      user = User.where(sso_guid: user_atts[:sso_guid]).first_or_initialize
```

knutsenm (changes requested):

app/services/auth_service_base.rb
andrewroth: I agreed, but the problem is test coverage. Seems really weird to be testing that it throws an error, but I can do it I guess.

app/services/google_auth_service.rb
andrewroth: Oh, yes, it will work now

knutsenm (approved):
🎉

frett:
@andrewroth this is good to merge, we just need to set the environment variables for production first

andrewroth:
@frett ok, I'll need a GOOGLE_APP_ID. Typically something like id.apps.googleusercontent.com.

andrewroth:
I have `APPLE_CLIENT_ID=org.cru.godtools` also, I believe that will be valid for production..

PR #1270, opened 2023-04-24
GT-1510 Include web only images when publishing a tool

frett:
- we no longer need to prevent web resources from being published
- add the web only images to the publish spec


frett:
the failing spec seems flaky. it's unrelated to my change, and running specs locally failed several times, but also succeeded a few times.

frett:
> ```
> 2023-04-26T15:09:49.2324491Z        expected: [{"id"=>"1", "type"=>"resource"}, {"id"=>"5", "type"=>"resource"}]
> 2023-04-26T15:09:49.2325002Z             got: [{"id"=>"5", "type"=>"resource"}, {"id"=>"1", "type"=>"resource"}]
> 2023-04-26T15:09:49.2325692Z 
> 2023-04-26T15:09:49.2325965Z        (compared using ==)
> ```
> 
> Does order matter here? If not, there are `match_array` and `contain_exactly` tests for "set equivalence"; see other Rails apps.

Order does not matter, I'll take a shot at updating the spec :)

knutsenm:
```
+++ b/spec/acceptance/users_controller_spec.rb
@@ -86,7 +86,7 @@
         expect(json_response["id"]).to eq(user.id.to_s)
         expect(json_response["type"]).to eq("user")
         expect(json_response.dig("relationships", "favorite-tools", "data").class).to eq(Array)
-        expect(json_response["relationships"]["favorite-tools"]["data"]).to eq([{"type" => "resource", "id" => resource.id.to_s}, {"type" => "resource", "id" => resource2.id.to_s}])
+        expect(json_response["relationships"]["favorite-tools"]["data"]).to match_array([{"type" => "resource", "id" => resource.id.to_s}, {"type" => "resource", "id" => resource2.id.to_s}])
       end
```

db/seeds.rb
```diff
@@ -133,6 +133,8 @@ def url
   Attachment.create!(resource: kgp, file: Rack::Test::UploadedFile.new("spec/fixtures/mobile_only.png", "image/png"), is_zipped: true)
   Attachment.create!(resource: kgp, file: Rack::Test::UploadedFile.new("spec/fixtures/web_mobile.png", "image/png"))
   Attachment.create!(resource: kgp, file: Rack::Test::UploadedFile.new("spec/fixtures/both.png", "image/png"), is_zipped: true)
+  Attachment.create!(resource: kgp, file: Rack::Test::UploadedFile.new("spec/fixtures/web_attach.png", "image/png"))
+  Attachment.create!(resource: kgp, file: Rack::Test::UploadedFile.new("spec/fixtures/web_bundled.png", "image/png"))
```
knutsenm, line 137: This doesn't seem to work for me locally, because line 130 fails

frett: would using `find_or_create_by` also work? testing it out locally right now

knutsenm: Hmm, no... then I get an error on line 131 because that particular file already exists

frett: ok, I get why it's failing, let me try out a couple things. I think just changing it to check for the last file might lead to some other conflicts as it creates all the files again. (not sure if rails would protected against that)

knutsenm (approved):

db/seeds.rb
knutsenm: This is not a problem in the Action where the test db is created from scratch

frett: I was using `bundle exec rails db:reset` locally when testing, which was why I didn't see the issue of migrating the old seed db to the new seed db

PR #1278, opened 2023-05-03
Remove the AppsFlyer analytics system from the content spec

frett:
@aaronlaib I'm removing the AppsFlyer analytics system due to us no longer using AppsFlyer. This won't break existing tools, but will prevent you from saving updates to a page that contain AppsFlyer analytics events.

### DIFF
--- a/public/xmlns/analytics.xsd
+++ b/public/xmlns/analytics.xsd
@@ -13,11 +13,6 @@
         <xs:list>
             <xs:simpleType>
                 <xs:restriction base="xs:token">
-                    <xs:enumeration value="appsflyer">
-                        <xs:annotation>
-                            <xs:documentation>AppsFlyer</xs:documentation>
-                        </xs:annotation>
-                    </xs:enumeration>
                     <xs:enumeration value="facebook">
                         <xs:annotation>
                             <xs:documentation>Facebook Analytics</xs:documentation>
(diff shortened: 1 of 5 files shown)


aaronlaib:
> @aaronlaib I'm removing the AppsFlyer analytics system due to us no longer using AppsFlyer. This won't break existing tools, but will prevent you from saving updates to a page that contain AppsFlyer analytics events.

Thanks @frett , looks good.

In regards to the XML, you mention this won't break anything, but will we lose any functionality until pages are updated? Looking at the test code you changed, I assume we'll want to change **appsflyer** to **firebase** in the example below? Same for the first pages of tracts where we've had appsflyer? Until those are updated, will we not have analytics for those pages?

`<analytics:events>
          <analytics:event system="appsflyer" action="lessonhq-0" />
 </analytics:events> `

frett:
@aaronlaib you won't lose functionality if tools aren't updated.

The few updates to the various schema test xmls were switched to other analytics systems to continue to test out including analytics events. In actual use you will probably just want to remove the appsflyer specific analytics events and not send them to other analytics systems.

aaronlaib:
> @aaronlaib you won't lose functionality if tools aren't updated.
> 
> The few updates to the various schema test xmls were switched to other analytics systems to continue to test out including analytics events. In actual use you will probably just want to remove the appsflyer specific analytics events and not send them to other analytics systems.
> 
> The reason we usually had the AppsFlyer event on the first page of the tool was because we weren't sending the default page analytics events to AppsFlyer, so we needed to explicitly send the few events we cared about. Firebase has always received the current page analytics event via the normal mechanism.

Sounds good, that is helpful to know.

aaronlaib (approved):

PR #1306, opened 2023-06-05
GT-1838, implement new apple login method with refresh tokens

andrewroth:


.env
```diff
@@ -41,4 +41,8 @@ FACEBOOK_APP_SECRET=facebook_app_secret
 
 GOOGLE_APP_ID=id.apps.googleusercontent.com
 
+APPLE_PRIVATE_KEY=private_key
 APPLE_CLIENT_ID=org.cru.godtools
```
levieggertcru, line 45: Hey @andrewroth will APPLE_CLIENT_ID and APPLE_REDIRECT_URI need to be dynamic here depending on ```staging``` vs ```production``` environment?  Or is PR for production only?

frett: @levieggertcru for our rails apps the environment variables are managed via a `secrets` command line utility and actually get set as environment variables in the Docker container when it is launched in our AWS infrastructure

levieggertcru: Thanks @frett makes sense.

andrewroth: Right, what @frett said :)

spec/fixtures/apple_verify_auth_code_response.json
```diff
@@ -0,0 +1 @@
+{"access_token":"aca3e04bbffa14ba1bbbdefe76c4afdb6.0.mrtwr.qKIemVfidy9dfzOBeQrazg","token_type":"Bearer","expires_in":3600,"refresh_token":"r10538d56ac494294936ebe0a3cea2b8b.0.mrtwr.TwQduM7CU-cCK5AGYj-Jdw","id_token":"eyJraWQiOiJZdXlYb1kiLCJhbGciOiJSUzI1NiJ9.eyJpc3MiOiJodHRwczovL2FwcGxlaWQuYXBwbGUuY29tIiwiYXVkIjoib3JnLmNydS5nb2R0b29scyIsImV4cCI6MTY4NjA2NzQyMywiaWF0IjoxNjg1OTgxMDIzLCJzdWIiOiIwMDEzNjEuYTVjYWZiN2Y0MmM4NDViODgwOWM0OGQwZjJiMDA4ODkuMTgwNCIsImF0X2hhc2giOiJmTjR1T1d3dG5GUXFkMmw5UEpKdklnIiwiZW1haWwiOiJsZXZpLmVnZ2VydEBnbWFpbC5jb20iLCJlbWFpbF92ZXJpZmllZCI6InRydWUiLCJhdXRoX3RpbWUiOjE2ODU5ODEwMDEsIm5vbmNlX3N1cHBvcnRlZCI6dHJ1ZX0.z-BCRqNqLO-jf_0ya8zczCLcA9Ukn2erV6-47RwWfr3gfZalIWXPV0RWH4e1CXYkAgYq6RrRv39Dyp62g_ddETV0jnvBLOrhzJ4-dQGJAWXwOonDM2BV0PMF1Tu_xiQnbcb12iZQ2Gom228krXowy4PcdCw4WRLGvRgSYESazuOrh3Ihvy3B7yifUb7Rs7gE4hpP4unonKfwxv7s_RzUasrAFYI4gvUjzfRryOGGBK1uR4hnQXl0Hc0oSkQv-01bknhYUtx3QsEFCF7J8v4P81eNsbl10qRa5zz3P2DOr8Fw8RHZtsduvUIZqn-WaCOWiBYzwXAQfLMY_5oaSZ6SIA"}
```
frett, line 1: just double checking that the refresh_token used in this sample is no longer actually valid with Apple? We don't want to accidentally leak a refresh_token that is effectively a credential for the mobile app

andrewroth: Good call. I've invalidated it now:

```
{"error":"invalid_grant","error_description":"The token has expired or has been revoked."}
```

frett (commented):
I'm good with this

I'll let @levieggertcru give input as to whether or not this is working functionally for iOS, and let Mark leave input on the ruby code itself

levieggertcru:
@frett @knutsenm  Staging is working.  Just tested and passed.

app/services/apple_auth_service.rb
```diff
@@ -2,22 +2,43 @@
...
+      validate_expected_fields!(id_token.raw_attributes)
+
+      id_token.verify!(
+        client: apple_id_client,
+        access_token: @response.access_token
+      )
 
-      apple_id_token = remote_user_id(decoded_token)
-      user_atts = extract_user_atts(apple_id_token, decoded_token, apple_id_token)
+      user_atts = {apple_user_id: id_token.sub, email: id_token.email}
```
knutsenm, line 26: Does `id_token.email` really exist? I don't see it in the docs at https://github.com/nov/apple_id/wiki/IdToken

frett: `email` is one of the [claims](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api/authenticating_users_with_sign_in_with_apple#3383773) in the `id_token` actually returned from apple

I would hope/assume that the `apple_id` library just exposes whatever claims get returned from apple.

Gemfile.lock
```diff
@@ -88,8 +79,14 @@ GEM
       jwt (>= 1.0, < 3.0)
       rails (>= 3.1)
       rest-client (>= 1.6.0)
+    aes_key_wrap (1.1.0)
     amazing_print (1.5.0)
+    apple_id (1.5.2)
```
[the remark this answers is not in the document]
andrewroth:
```
Fetching gem metadata from https://rubygems.org/.......
Resolving dependencies......
Bundler could not find compatible versions for gem "faraday":
  In snapshot (Gemfile.lock):
    faraday (= 1.10.3)

  In Gemfile:
    googleauth was resolved to 0.17.1, which depends on
      faraday (>= 0.17.3, < 2.0)

    apple_id (= 1.6.2) was resolved to 1.6.2, which depends on
      rack-oauth2 (~> 2.0) was resolved to 2.2.0, which depends on
        faraday (~> 2.0)

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.
```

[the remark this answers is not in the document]
andrewroth: Interesting, there's a newer googleauth as well, but it's not installing because of google-api-client, which is version-locked at ~> 0.53

```
Fetching gem metadata from https://rubygems.org/.......
Resolving dependencies......
Bundler could not find compatible versions for gem "googleauth":
  In Gemfile:
    googleauth (= 1.5.2)

    google-api-client (~> 0.53) was resolved to 0.53.0, which depends on
      google-apis-core (~> 0.1) was resolved to 0.2.0, which depends on
        googleauth (~> 0.14)
```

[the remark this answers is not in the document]
knutsenm: If you `bundle update apple_id googleauth google-api-client`, it will say
> Bundler attempted to update google-api-client but its version stayed the same
but will update apple_id anyway.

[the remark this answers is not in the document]
andrewroth: Ok, can you commit it, then?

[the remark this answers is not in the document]
andrewroth: And looks like google api client is used in `app/lib/update_global_activity_analytics.rb`. Looks like we're supposed to switch to the gem corresponding to the specific "service" we want: https://github.com/googleapis/google-api-ruby-client/commit/ded98b2afa1c59b73c0debe6565454120cad9546

knutsenm (approved):

frett:
if testing for coverage is too burdensome I'm good with relaxing that requirement to merge the PR.

I think the bigger requirement is whether the functionality is actually being well tested or not. I'll leave it up to your discretion as to where you land within that requirement.

andrewroth:
> if testing for coverage is too burdensome I'm good with relaxing that requirement to merge the PR.
> 
> I think the bigger requirement is whether the functionality is actually being well tested or not. I'll leave it up to your discretion as to where you land within that requirement.

I got it now, it's covered at 100%. Shall I merge?

frett:
yeah, go for it

frett:
did the environtment variables get set for production already?

PR #1326, opened 2023-07-11
[GT-1972] API - Tool Groups

cesarediaz:
This PR belongs to this Jira issue [https://jira.cru.org/browse/GT-1972](https://jira.cru.org/browse/GT-1972)

EndPoints

**Tool Groups**
- [x] POST /tool-groups (creates a new tool-group)
- [x] GET /tool-groups (retrieves a list of all current tool-groups)
- [x] GET /tool-groups/1 (retrieves tool-group 1)
- [x] PATCH /tool-groups/1 (update tool-group 1)
- [x] DELETE /tool-groups/1 (delete tool-group 1)

**Rule Languages**
- [x] POST /tool-groups/1/rules-language (create a new rule for tool-group 1)
- [x] PATCH /tool-groups/1/rule-language/1 (update rule 1 for tool-group 1)
- [x] DELETE /tool-groups/1/rule-language/1 (delete rule 1 for tool-group 1)

**Rules Country**
- [x] POST /tool-groups/1/rules-country (create a new rule for tool-group 1)
- [x] PATCH /tool-groups/1/rules-country/1 (update rule 1 for tool-group 1)
- [x] DELETE /tool-groups/1/rules-country/1 (delete rule 1 for tool-group 1)

**Rules Praxis**
- [x] POST /tool-groups/1/rules-praxis (create a new rule for tool-group 1)
- [x] PATCH /tool-groups/1/rules-praxis/1 (update rule 1 for tool-group 1)
- [x] DELETE /tool-groups/1/rules-praxis/1 (delete rule 1 for tool-group 1)

**Resources (Tool)**
- [x] POST /tool-groups/1/tools (create a new relationship to a resource (aka a tool) for tool-group 1)
- [x] PATCH /tool-groups/1/tools/1 (update relationship & attributes for that relationship)
- [x] DELETE /tool-groups/1/tools/1 (delete relationship)

**Suggestions**
- [x] GET /resources/suggestions


config/routes.rb
```diff
@@ -30,6 +30,8 @@
 
   resources :custom_manifests, only: [:create, :update, :destroy, :show]
 
+  resources :tool_groups, only: [:create, :destroy, :index, :show, :update]
```
frett, line 33: On stage the API is currently responding at `/tool_groups` and not `/tool-groups`, I think this will fix that

```suggestion
  resources :tool_groups, path: "tool-groups", only: [:create, :destroy, :index, :show, :update]
```

app/controllers/tool_groups_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+  before_action :authorize!
+
+  def index
+    render json: ToolGroup.all.order(name: :asc), status: :ok
+  end
+
+  def create
+    create_tool_group
+  rescue ActiveRecord::RecordInvalid
+    update_tool_group
```
frett, line 11: I don't think we want to rescue and update an existing record if creating it fails. we should return the error that occurred when trying to create it (e.g. missing attribute)

frett: I just encountered this issue when trying to create a tool-group with a mis-named attribute, the error I got back was that it couldn't find a tool group with the specified id, which was not the actual problem I had.

app/serializers/tool_group_serializer.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class ToolGroupSerializer < ActiveModel::Serializer
+  attributes :id, :name, :suggestions_weight
```
frett, line 4: the `_` in `suggestions_weight` isn't translating properly to "suggestions-weight" for the API endpoints, what needs to happen to fix it is to specify an explicit key for that attribute. see: https://github.com/CruGlobal/mobile-content-api/blob/master/app/serializers/resource_serializer.rb#L6

```suggestion
  attributes :id, :name
  attribute :suggestions_weight, key: "suggestions-weight"
```

app/controllers/tool_groups_controller.rb
cesarediaz: I was working in this when you commented it. I think now it is just returning the error message and not trying to rescue.

I'll mar it as resolved

app/serializers/tool_group_serializer.rb
cesarediaz: @frett the change is there

app/models/rule_praxi.rb
```diff
@@ -0,0 +1,24 @@
+class RulePraxi < ApplicationRecord
```
frett, line 1: so, praxis is the actual name, it's not a plural of praxi

```suggestion
class RulePraxis < ApplicationRecord
```

app/serializers/tool_group_serializer.rb
```diff
@@ -0,0 +1,12 @@
...
+
+class ToolGroupSerializer < ActiveModel::Serializer
+  attributes :id, :name
+  attribute :suggestions_weight, key: "suggestions-weight"
+
+  type "tool-group"
+
+  has_many :rule_languages
+  has_many :rule_countries
+  has_many :rule_praxis
```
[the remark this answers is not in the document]
frett: this naming should also be used for them in the routes

app/controllers/tool_groups_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+  before_action :authorize!
+
+  def index
+    render json: ToolGroup.all.order(name: :asc), status: :ok
+  end
+
+  def create
+    create_tool_group
+  rescue ActiveRecord::RecordInvalid => e
+    render json: {error: e.record.errors}, status: :unprocessable_entity
```
frett, line 11: so this error format isn't currently following the jsonapi spec, so our client parsers won't be able to parse it correctly.

relevant [examples](https://jsonapi.org/examples/#error-objects-basics) and the (spec)[https://jsonapi.org/format/#error-objects]

So, currently the errors are being returned like this:

```json
{
    "error": {
        "suggestions_weight": [
            "can't be blank"
        ],
        "name": [
            "has already been taken"
        ]
    }
}
```

When it should look more like:
```json
{
  "errors": [
    {
      "detail": "name has already been taken"
    },
    {
      "detail": "suggestions-weight can't be blank"
    }
  ]
}
```

There are other possible fields that can be included in the error objects, but the detail one is what we currently use on the admin UI to display any errors returned from the API

frett: It actually looks like if you raise an error, it will automatically be formatted as a jsonapi error correctly.

https://github.com/CruGlobal/mobile-content-api/blob/master/app/controllers/languages_controller.rb#L19

yields the following JSON:
```json
{
    "errors": [
        {
            "source": {
                "pointer": "/data/attributes/id"
            },
            "detail": "Code en already exists."
        }
    ]
}
```

config/routes.rb
```diff
@@ -30,6 +30,23 @@
 
   resources :custom_manifests, only: [:create, :update, :destroy, :show]
 
+  resources :tool_groups, path: "tool-groups", only: [:create, :destroy, :index, :show, :update]
+
+  # Rule Languages
+  post "tool-groups/:id/rule-languages", to: "rule_languages#create"
```
frett, line 36: the `:id` parameter here doesn't match the `:tool_group_id` parameter you are using within `rule_languages#create`, this yields a "tool_group must exist" error

app/controllers/tool_groups_controller.rb
```diff
@@ -0,0 +1,45 @@
+class ToolGroupsController < ApplicationController
+  before_action :authorize!
+
+  def index
+    render json: ToolGroup.all.order(name: :asc), status: :ok
```
frett, line 5: One of the aspects of jsonapi that we leverage is using the `include` parameter to dynamically including related objects we care about. In order for it to function we need to pass the parameter into the render logic.

```suggestion
    render json: ToolGroup.all.order(name: :asc), include: params[:include], status: :ok
```

frett: example from another controller: https://github.com/CruGlobal/mobile-content-api/blob/master/app/controllers/languages_controller.rb#L11

we should also include the fields param to be thorough.

frett: the include parameter only needs to be supported for the `index` and `show` methods

app/controllers/tool_groups_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def create
```
frett, line 8: when making this call:
```
curl --location 'https://mobile-content-api-stage.cru.org/tool-groups' \
--header 'Accept: application/vnd.api+json' \
--header 'Content-Type: application/vnd.api+json' \
--data '{
 "data": {
  "type": "tool-group",
  "attributes": {
   "name": "Tool Group Name 3",
   "suggestions-weight": "1.0"
  }
 }
}'
```
I get an error "suggestions_weight can't be blank".

is it possible to have it be consistent with how we spell `suggestions-weight` elsewhere using the `-`? I'm not sure if we've ever actually made this consistent elsewhere within the API before. maybe just do some standardization of attribute names (`tr('-', '_')`) when they are being parsed by the API to create/update a data model?

frett: we can discuss this if it'll be too convoluted to standardize it here.

The higher importance for me in consistency for attribute names/relationship names & paths across the API

app/serializers/tool_group_serializer.rb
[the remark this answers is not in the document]
cesarediaz: fixed now

app/controllers/tool_groups_controller.rb
frett: maybe this logic can be extracted to a common `ToolGroupsController#permitted_params` method since it will be used for both creates and updates.

e.g. https://github.com/CruGlobal/mobile-content-api/blob/master/app/controllers/resources_controller.rb#L72-L74

app/controllers/tool_groups_controller.rb
frett: it looks like there is some common methods for rendering jsonapi errors that can probably be utilized:

https://github.com/CruGlobal/mobile-content-api/blob/master/app/controllers/application_controller.rb#L68-L86

cesarediaz: It should be fixed now

app/controllers/tool_groups_controller.rb
```diff
@@ -2,7 +2,13 @@ class ToolGroupsController < ApplicationController
   before_action :authorize!
 
   def index
-    render json: tool_groups_ordered_by_name, status: :ok
+    include = {}
+    fields = {}
+
+    include = params[:include]&.split(",") if params[:include]
+    fields = get_fields(params[:fields]) if params[:fields]
```
frett, line 9: I'm guessing you could use the [`fields_param`](https://github.com/CruGlobal/mobile-content-api/blob/master/app/controllers/concerns/fields.rb#L4) function here and not need to implement a separate get_fields() function.

I haven't looked into the pre-existing vs your implementation to know which one would be more accurate/safe

app/controllers/tool_groups_controller.rb
```diff
@@ -2,7 +2,13 @@ class ToolGroupsController < ApplicationController
   before_action :authorize!
 
   def index
-    render json: tool_groups_ordered_by_name, status: :ok
+    include = {}
+    fields = {}
+
+    include = params[:include]&.split(",") if params[:include]
```
frett, line 8: I'm surprised you need to manually split on the `,`, we haven't had to do that with our other controllers and regularly request multiple included relationships from those endpoints

cesarediaz: You are rigth, running tests without this split there works fine

app/controllers/tool_groups_controller.rb
cesarediaz: That is right! I think the problem was just in the way the params were passed in the specs and that was not working with the  `fields_param` . Now it works well with the pre existent `fields_param` function

config/routes.rb
cesarediaz: I think it is solved when sent code in commit [0ee2a267b7b03c7a9480b4f865809dddd33b5413](https://github.com/CruGlobal/mobile-content-api/pull/1326/commits/0ee2a267b7b03c7a9480b4f865809dddd33b5413#diff-959bc9abc46a55332bb64d5155a79323afa75a50ec1a2137ddd22d926f62c6c5R36)

app/controllers/tool_groups_controller.rb
cesarediaz: Daniel, I've implemented a [method](https://github.com/CruGlobal/mobile-content-api/commit/435a5ba6a8e60fc53daeffd6652d2943a6fdd0bd#diff-766c34fd6533171eaf54300c153f89d6002c35c02cfc9c5b219251f85180ad07R123) that is available for all controllers and converts the hyphens into underscore.

Now we can do a POST request like this

```
curl --location 'http://localhost:3000/tool-groups' \
--header 'Content-Type: application/vnd.api+json' \
--header 'Accept: application/vnd.api+json' \
--data '{
 "data": {
  "type": "tool-group",
  "attributes": {
   "name": "Test Group Name 10",
   "suggestions-weight": "1.0"
  }
 }
}'
```

or  a PUT request like this

```
curl -X PUT --location 'http://localhost:3000/tool-groups/10' \
--header 'Content-Type: application/vnd.api+json' \
--header 'Accept: application/vnd.api+json' \
--data '{
 "data": {
  "type": "tool-group",
  "attributes": {
   "name": "Test Group Name 10.1",
   "suggestions-weight": "1.1"
  }
 }
}'
```

frett: cool, I'll get back to testing this shortly, in the middle of a couple other tasks

app/models/rule_praxi.rb
cesarediaz: I think this should be following the RoR inflection pattern now with the `RulePraxis` model name

app/controllers/rule_languages_controller.rb
```diff
@@ -0,0 +1,35 @@
+class RuleLanguagesController < ApplicationController
+  before_action :authorize!
+
+  def create
+    create_rule_language
```
frett, line 5: when issuing the follow create request:

```
curl --location 'https://mobile-content-api-stage.cru.org/tool-groups/1/rules-language' \
--header 'Content-Type: application/vnd.api+json' \
--header 'Accept: application/vnd.api+json' \
--data '{
 "data":   {
   "type": "tool-group-rule-language",
   "attributes": {
    "languages": [
     "en"
    ]
   }
 }
}'
```

I get the following response:
```
{
    "error": {
        "tool_group": [
            "must exist"
        ]
    }
}
```

frett: 2 things wrong here:
1. it's not picking up the tool-group from the tool-group id in the URL
2. the error format isn't using the jsonapi error format

app/controllers/rule_languages_controller.rb
```diff
@@ -0,0 +1,35 @@
+class RuleLanguagesController < ApplicationController
+  before_action :authorize!
+
```
frett, line 3: you will need to add the convert dashes to hyphens `before_action` here.

The following request doesn't update the `negative-rule` attribute.
```
curl --location --request PUT 'https://mobile-content-api-stage.cru.org/tool-groups/1/rules-language/1' \
--header 'Content-Type: application/vnd.api+json' \
--header 'Accept: application/vnd.api+json' \
--data '{
 "data":   {
   "type": "tool-group-rule-language",
   "attributes": {
       "negative-rule": false
   }
 }
}'
```

app/controllers/rule_languages_controller.rb
cesarediaz: It was solved in last commits from yesterday

app/controllers/rule_languages_controller.rb
cesarediaz: Solved in commits yesterday

app/serializers/resource_serializer.rb
```diff
@@ -16,6 +16,10 @@ class ResourceSerializer < ActiveModel::Serializer
   has_many :custom_manifests, key: "custom-manifests"
   has_many :variants, if: -> { object&.resource_type&.name == "metatool" }
   has_many :translated_attributes, key: "translated-attributes"
+
+  has_many :resource_tool_groups
+  has_many :tool_groups, through: :resource_tool_groups
+
```
frett, line 22: we don't need to expose this via the resource serializer

app/serializers/tool_group_serializer.rb
```diff
@@ -0,0 +1,26 @@
...
+  attributes :id, :name
+  attribute :suggestions_weight, key: "suggestions-weight"
+
+  type "tool-group"
+
+  has_many :rule_languages, key: "rules-language"
+  has_many :rule_countries, key: "rules-country"
+  has_many :rule_praxes, key: "rules-praxis"
+  has_many :resource_tool_groups, key: "resource-tool-groups"
+  has_many :resources, through: :resource_tool_groups
```
frett, line 13: we don't need to expose this directly through the serializer, exposing the intermediary relationship is enough

app/serializers/tool_group_serializer.rb
```diff
@@ -0,0 +1,26 @@
...
+class ToolGroupSerializer < ActiveModel::Serializer
+  attributes :id, :name
+  attribute :suggestions_weight, key: "suggestions-weight"
+
+  type "tool-group"
+
+  has_many :rule_languages, key: "rules-language"
+  has_many :rule_countries, key: "rules-country"
+  has_many :rule_praxes, key: "rules-praxis"
+  has_many :resource_tool_groups, key: "resource-tool-groups"
```
frett, line 12: use a key of `tools` for this

```suggestion
  has_many :resource_tool_groups, key: "tools"
```

app/controllers/rule_countries_controller.rb
```diff
@@ -0,0 +1,36 @@
+class RuleCountriesController < ApplicationController
+  before_action :authorize!
+
```
frett, line 3: this needs to the dash to hyphen's `before_action`, just verified that `negative-rule` is not being set/updated correctly

app/serializers/tool_group_serializer.rb
frett: @cesarediaz there are still a few outstanding change requests on the tool-groups APIs

cesarediaz: solved

app/controllers/rule_countries_controller.rb
cesarediaz: applied

app/controllers/tool_groups_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+  def create_tool
+    ResourceToolGroup.create!(
+      tool_group_id: params[:tool_group_id],
+      resource_id: params[:data][:attributes]["resource-id"],
+      suggestions_weight: params[:data][:attributes]["suggestions-weight"]
+    )
+
+    tool_group = ToolGroup.find(params[:tool_group_id])
+    response.headers["Location"] = "tool_groups/#{tool_group.id}"
+    render json: tool_group, status: :created
```
frett, line 24: we actually should be rendering the ResourceToolGroup here, again I'll work up an example for this API endpoint

app/controllers/tool_groups_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+  def create
+    create_tool_group
+  rescue ActiveRecord::RecordInvalid => e
+    render json: {errors: formatted_errors(e)}, status: :unprocessable_entity
+  end
+
+  def create_tool
+    ResourceToolGroup.create!(
+      tool_group_id: params[:tool_group_id],
+      resource_id: params[:data][:attributes]["resource-id"],
```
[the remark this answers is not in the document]
frett: sample request to create the relationship:
```
curl --location 'https://mobile-content-api-stage.cru.org/tool-groups/1/tools' \
--header 'Content-Type: application/vnd.api+json' \
--header 'Accept: application/vnd.api+json' \
--data '{
 "data":   {
   "type": "tool-group-tool",
   "attributes": {
    "suggestions-weight": "1.0"
   },
   "relationships": {
    "tool": {
     "data": {
      "type": "resource",
      "id": 1
     }
    }
   }
 }
}'
```

app/controllers/tool_groups_controller.rb
frett: I would expect the response to look like:
```
{
   "id": 1,
   "type": "tool-group-tool",
   "attributes": {
    "suggestions-weight": "0.1"
   },
   "relationships": {
    "tool-group": {
     "data": {
      "id": 1,
      "type": "tool-group"
     }
    },
    "tool" : {
     "data": {
      "id": 3,
      "type": "resource"
     }
    }
   }
```

app/serializers/resource_tool_group_serializer.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class ResourceToolGroupSerializer < ActiveModel::Serializer
+  attributes :resource_id, :tool_group
```
frett, line 4: don't include the `resource_id` or `tool_group` here

app/serializers/resource_tool_group_serializer.rb
```diff
@@ -0,0 +1,11 @@
...
+  attribute :suggestions_weight, key: "suggestions-weight"
+
+  type "tool-group-tool"
+
+  belongs_to :resource
```
frett, line 9:
```suggestion
  belongs_to :resource, key: "tool"
```

app/controllers/rule_praxes_controller.rb
```diff
@@ -0,0 +1,36 @@
+class RulePraxesController < ApplicationController
+  before_action :authorize!
```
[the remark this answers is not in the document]
cesarediaz: implemented

app/serializers/resource_tool_group_serializer.rb
cesarediaz: excluded the attributes :resource_id & :tool_group

app/serializers/resource_tool_group_serializer.rb
cesarediaz: change made

app/controllers/resources_controller.rb
```diff
@@ -34,8 +34,76 @@ def push_to_onesky
...
+          resource_id: resource.resource_id,
+          resource_tool_group_suggestions_weight: resource.suggestions_weight
+        }
+      end
+    end
+
+    result.group_by { |hash| hash[:resource_id] }
+  end
+
+  def match_params(tool_group, params)
```
frett, line 73: I created a tool group with no defined rules and a single tool in it, but that tool is not being returned by the `/resources/suggestions` endpoint. Since the group had no rules it should be considered a match for any suggestion request.

This would probably also affect having only a single rule defined for a tool group, we will probably define a tool-group for a specific language only and not restrict it by country or the praxis scale, in which case that group should match as long as the specific language matches and not care about the rest of the defined filters

frett: or similarly there is a good chance we will define groups with just a country rule or just a praxis rule

spec/acceptance/tool_group_controller_spec.rb
```diff
@@ -0,0 +1,312 @@
...
+      expect(status).to eq(400)
+      expect(JSON.parse(response_body)["errors"]).not_to be_empty
+      expect(JSON.parse(response_body)["errors"][0]["detail"]).to eql "Validation failed: Suggestions weight can't be blank"
+    end
+  end
+
+  post "tool-groups/:tool_group_id/tools" do
+    let(:attrs) do
+      {
+        "relationships": {
```
frett, line 73: `relationships` shouldn't be a child of `attributes`, it should be at the same level as attributes.

```
{
  "attributes": {},
  "relationships": {}
}
```

app/controllers/tool_groups_controller.rb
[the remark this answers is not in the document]
cesarediaz: this should be solved now

spec/acceptance/tool_group_controller_spec.rb
cesarediaz: solved

app/controllers/resources_controller.rb
cesarediaz: Ok, I think understood the idea. I'll work on tests for those cases and let you know when I've a case to test

cesarediaz: @frett I've added code for return as valid tool group when it has not rule/s

cesarediaz: The second part is ...

_"This would probably also affect having only a single rule defined for a tool group, we will probably define a tool-group for a specific language only and not restrict it by country or the praxis scale, in which case that group should match as long as the specific language matches and not care about the rest of the defined filters"_

@frett , so means that if a group tool has just one rule and it is about a "language rule" with a single language(not more than one language), this "tool group" should be considered as a match automatically without taking care of other param?

app/serializers/tool_group_serializer.rb
frett: @cesarediaz this still needs to be adjusted, but it's a simple change :)

app/services/tool_filter_service.rb
```diff
@@ -0,0 +1,102 @@
...
+  end
+
+  # Returns true if the tool group has a single rule with a language/s
+  # present in the 'languages' array, otherwise returns false.
+  def language_rule(tool_group, languages)
+    return false unless tool_group.rule_languages.one?
+    return false unless tool_group.rule_praxes.none? || tool_group.rule_countries.none?
+
+    rule_languages = tool_group.rule_languages.first.languages
+    (rule_languages & languages).any?
```
frett, line 100: TypeError: no implicit conversion of nil into Array

https://app.rollbar.com/a/Cru/fix/item/mobile-content-api/320

frett:
```
curl --location --globoff 'https://mobile-content-api-stage.cru.org/resources/suggestions?filter[country]=FR'
```

app/services/tool_filter_service.rb
```diff
@@ -0,0 +1,102 @@
...
+          resource_tool_group_suggestions_weight: resource.suggestions_weight
+        }
+      end
+    end
+
+    result.group_by { |hash| hash[:resource_id] }
+  end
+
+  def match_params(tool_group, params)
+    country = params["filter"]["country"]&.upcase
```
frett, line 44: NoMethodError: undefined method `[]' for nil:NilClass

https://app.rollbar.com/a/Cru/fix/item/mobile-content-api/319

```
curl --location 'https://mobile-content-api-stage.cru.org/resources/suggestions'
```

app/services/tool_filter_service.rb
```diff
@@ -0,0 +1,102 @@
...
+    languages = params["filter"]["languages"]
+    openness = params["filter"]["openness"].to_i
+    confidence = params["filter"]["confidence"].to_i
+
+    return true if no_rules_for(tool_group)
+    return true if language_rule(tool_group, languages)
```
frett, line 50: I have a tool group defined with the following 2 rules:
- language: en
- praxis: openness: 1 confidence: 3

When i submit the following suggestions query:

```
curl --location --globoff 'https://mobile-content-api-stage.cru.org/resources/suggestions?filter[languages][]=en&filter[openness]=1&filter[confidence]=1'
```

I would expect to not get back any results because the confidence doesn't match, but it is still matching the group and returning results.

frett: I believe this logic is where it is failing because `language_rule` has the following statement:
```
return false unless tool_group.rule_praxes.none? || tool_group.rule_countries.none?
```
which still processes as true because there were no defined country rules.

Also, since you have each rule type only being evaluated if a rule exists you probably no longer need the `no_rules_for` or `language_rule` special cases.

app/services/tool_filter_service.rb
cesarediaz: I'll revise this now

app/services/tool_filter_service.rb
frett: Screenshot 2023-08-01 at 4 56 40 PM

app/services/tool_filter_service.rb
cesarediaz: Take a look please, I think I have solved the problem

app/controllers/tool_groups_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+    existing = resource_tool_group
+    existing.update!(
+      resource_id: params[:data][:attributes]["resource-id"],
+      suggestions_weight: params[:data][:attributes]["suggestions-weight"]
+    )
+    render json: existing, status: :accepted
+  end
+
+  def delete_tool
+    resource_tool_group.destroy!
```
frett, line 42: NoMethodError: undefined method `destroy!' for nil:NilClass

https://app.rollbar.com/a/Cru/fix/item/mobile-content-api/322

Screenshot 2023-08-02 at 2 20 08 PM

frett: this happens when trying to delete a relationship that doesn't exist

frett: also, it appears that this logic ignore which tool group the relationship is being deleted from, so I can delete a relationship from a different tool group.

Scenario:
- tool group: 1 relationship id: 1
- tool group: 2 relationship id: 2
- The following delete request should fail: `DELETE /tool-groups/2/tools/1` because relationship 1 doesn't exist for tool-group 2
   - Currently this request will succeed because it doesn't validate that the requested relationship is actually part of the specified tool-group

I haven't verified this, but this same behavior is probably missing from `update_tool` as well because it also uses the same `resource_tool_group` function.

app/controllers/tool_groups_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+    head :no_content
+  end
+
+  def show
+    render json: load_tool_group, include: params[:include], fields: field_params, status: :ok
+  end
+
+  def destroy
+    tool_group = ToolGroup.find(params[:id])
+    tool_group.destroy!
```
frett, line 52: > ActiveRecord::InvalidForeignKey: PG::ForeignKeyViolation: ERROR:  update or delete on table "tool_groups" violates foreign key constraint "fk_rails_c56fad9d31" on table "resource_tool_groups"
DETAIL:  Key (id)=(4) is still referenced from table "resource_

https://app.rollbar.com/a/Cru/fix/item/mobile-content-api/323

Screenshot 2023-08-02 at 2 29 18 PM

frett: deleting a tool-group that still contains relationships to `resources` should succeed and delete those relationships.

it should NOT cascade and delete the `resources` themselves, just the relationships

app/controllers/tool_groups_controller.rb
cesarediaz: should it return a message error or just not explode?

app/controllers/tool_groups_controller.rb
cesarediaz: should be fixed now

app/controllers/tool_groups_controller.rb
cesarediaz: fixed

app/services/tool_filter_service.rb
```diff
@@ -0,0 +1,114 @@
...
+      end
+    end
+
+    return true if no_rules_for(tool_group)
+
+    # Rule Countries
+    if tool_group.rule_countries.any?
+      if country
+        if tool_group.rule_countries.any? { |o| o.countries.exclude?(country) && o.negative_rule }
+          return true
```
frett, line 65: a tool-group defined with the following 2 rules:
- languages: de
- countries: FR

when issuing the following query the group incorrectly matches:
`https://mobile-content-api-stage.cru.org/resources/suggestions?filter[country]=FR`

The group shouldn't match because the languages rule was not satisfied by the provided filters

frett: I think this initial `if` statement can go away and just the 2 `elsifs` are needed

cesarediaz: Both rules are `negative_rule: false ` ?

cesarediaz: It should be solved now

app/services/tool_filter_service.rb
```diff
@@ -0,0 +1,106 @@
...
+    # Rule Countries
+    if tool_group.rule_countries.any? && !country.nil?
+      if tool_group.rule_countries.any? { |o| o.countries.include?(country) && o.negative_rule }
+        return false
+      elsif tool_group.rule_countries.any? { |o| o.countries.exclude?(country) && !o.negative_rule }
+        return false
+      end
+    elsif !tool_group.rule_countries.any? && !country.nil? && !tool_group.rule_languages.any?
+      return false
+    end
```
frett, line 66:
```suggestion
    if tool_group.rule_countries.any?
      if tool_group.rule_countries.any? { |o| o.negative_rule && o.countries.include?(country) }
        return false
      elsif tool_group.rule_countries.any? { |o| !o.negative_rule && o.countries.exclude?(country) }
        return false
      end
    end
```

frett: @cesarediaz this is passing all the specs, I would guess this is all we would need for country rules. Similar changes can be made to languages, openness, and confidence rule checks

app/services/tool_filter_service.rb
```diff
@@ -0,0 +1,106 @@
...
+    # Rule Languages
+    if tool_group.rule_languages.any? && languages
+      if tool_group.rule_languages.any? { |o| (languages & tool_group.rule_languages.first.languages).empty? && !o.negative_rule }
+        return false
+      elsif tool_group.rule_languages.any? { |o| (languages - tool_group.rule_languages.first.languages).empty? && o.negative_rule }
+        return false
+      end
+    elsif tool_group.rule_languages.any? && !languages.nil?
+      return false
+    end
```
frett, line 77:
```suggestion
    # Language Rules
    if tool_group.rule_languages.any? { |o| !o.negative_rule && (languages & o.languages).empty? }
      return false
    elsif tool_group.rule_languages.any? { |o| o.negative_rule && (languages - o.languages).empty? }
      return false
    end
```

app/controllers/resources_controller.rb
```diff
@@ -34,6 +34,10 @@ def push_to_onesky
     head :no_content
   end
 
+  def suggestions
+    render json: ToolFilterService.new(params).call, status: :ok
```
frett, line 38: I believe we should be able to include `include_param: params[:include], fields_param: field_params` when rendering this json to be able to support any includes/fields for the response

frett: I'm not totally sure if that will be beneficial or not

cesarediaz: ok, I'll work on it

app/controllers/tool_groups_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+    existing = resource_tool_group
+    existing.update!(
+      resource_id: params[:data][:attributes]["resource-id"],
+      suggestions_weight: params[:data][:attributes]["suggestions-weight"]
+    )
+    render json: existing, status: :accepted
+  end
+
+  def delete_tool
+    if resource_tool_group&.destroy!
```
frett, line 42: when looking up the `resource_tool_group`, this is ignoring the tool-group id. This means the following request succeeds when it should have returned a 404 instead.

```
curl --location --request DELETE 'https://mobile-content-api-stage.cru.org/tool-groups/invalid/tools/2' \
--header 'Accept: application/vnd.api+json' \
```

app/controllers/tool_groups_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+    response.headers["Location"] = "tool-groups/#{resource_tool_group.id}"
+    render json: resource_tool_group, status: :created
+  rescue ActiveRecord::RecordInvalid => e
+    render json: {errors: formatted_errors(e)}, status: :unprocessable_entity
+  end
+
+  def update_tool
+    existing = resource_tool_group
+    existing.update!(
+      resource_id: params[:data][:attributes]["resource-id"],
```
frett, line 35: the resource id should be read from the relationships json.

```
curl --location --request PUT 'https://mobile-content-api-stage.cru.org/tool-groups/1/tools/5' \
--header 'Content-Type: application/vnd.api+json' \
--header 'Accept: application/vnd.api+json' \
--header 'Authorization: eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VyX2lkIjoxLCJleHAiOjE2OTExODE4MDUsImlhdCI6MTY5MTA5NTQwNX0.UVBAMJg0Gq8eX3taz9I1REJevJER05y4DNo-mInhaVM' \
--data '{
 "data":   {
   "type": "tool-group-tool",
   "attributes": {
    "suggestions-weight": "1.0"
   },
   "relationships": {
    "tool": {
     "data": {
      "type": "resource",
      "id": 1
     }
    }
   }
 }
}'
```

app/controllers/resources_controller.rb
cesarediaz: can you give me some curl request sample for this?

frett:
```
https://mobile-content-api-stage.cru.org/resources/suggestions?filter[country]=AU&filter[confidence]=1&include=latest-translations&fields[resource]=name%2Cabbreviation%2Clatest-translations
```

frett: this is working correctly

app/controllers/tool_groups_controller.rb
cesarediaz: this should be solved now

app/controllers/tool_groups_controller.rb
cesarediaz: this should be solved now too

db/schema.rb
```diff
@@ -160,8 +160,6 @@
     t.string "key", null: false
     t.string "value", null: false
     t.boolean "is_translatable", default: false
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
```
andrewroth, line 164: Was this intentional? I don't think these should be deleted unless it was intentional and there's a reason for it.

spec/acceptance/resources_controller_spec.rb
```diff
@@ -9,6 +9,421 @@
...
+          expect(JSON.parse(response_body)["data"].count).to eql 0
+        end
+      end
+
+      context "if matching any of the languages defined in rule" do
+        it "return coincidences" do
+          do_request "filter[language]": languages_fr_it
+
+          expect(status).to be(200)
+          expect(JSON.parse(response_body)["data"].count).to eql 5
```
andrewroth, line 76: can the coincidences be checked that they're the correct ones and not just that there's 5?

andrewroth: I'm actually kind of surprise and confused to see the term "coincidence" -- which is defined as "a remarkable concurrence of events or circumstances without apparent causal connection." -- is that what was intended? Or is this maybe "co" "incidence", like a match? overlap? maybe a better term would be less confusing?

Nowhere in the code above is the term "coincidence" used.

andrewroth (changes requested):
Looks pretty good overall.

I woud like to see the coverage get a green check, there's some places in the Files changed view you can see that Codecov is pointing out is not covered.

db/schema.rb
```diff
@@ -182,6 +180,16 @@
     t.index ["resource_id"], name: "index_pages_on_resource_id"
   end
 
+  create_table "resource_tool_groups", force: :cascade do |t|
+    t.bigint "resource_id", null: false
```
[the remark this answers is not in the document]
knutsenm: The docs say,

> The bigint type is designed to be used when the range of the integer type is insufficient

and

> The type integer is the common choice, as it offers the best balance between range, storage size, and performance

So, if you don't need to store numbers > 2B, stick with `t.integer`.

On the other hand, this merely follows suit from the existing `resource_id` in `language_attributes` - but I question the need for that bigint as well. These are all foreign keys to IDs that are only int4 - plain integer.

spec/acceptance/resources_controller_spec.rb
knutsenm: "coincident" means "occurring together in space or time" - perhaps that is what was meant - or "coinciding"

db/schema.rb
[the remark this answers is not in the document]
cesarediaz: Done

db/schema.rb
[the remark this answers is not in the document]
cesarediaz: @andrewroth @knutsenm 

I think that this is a default behaviour of RoR migrations, although ofcourse we can convertet them into Integer.

If we don't need this bigint, should we consider to convert to this to `Integer` type in both tables `resource_tool_groups` and `language_attributes` column `resource_id`?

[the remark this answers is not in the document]
andrewroth: That sounds good to me.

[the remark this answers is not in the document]
cesarediaz: @andrewroth  
Good, then I will convert column type for both tables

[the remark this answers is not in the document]
knutsenm: > I think that this is a default behaviour of RoR migrations

I doubt it, but I'm not positive about that. In production, resources.id is an `int4` - i.e. a plain `integer`, so the `t.references :resource` should have generated an `integer` here.

Perhaps your local database has a bigint for `resources.id` ? Which version of PostgreSQL are you running? We're on v11 (which surprises me - we've moved to v14 most everywhere)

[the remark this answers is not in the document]
cesarediaz: Done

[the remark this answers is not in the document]
knutsenm: You were right: Rails defaults to bigint for new primary keys since 5.1: https://github.com/rails/rails/pull/26266

That doesn't explain, however, why this reference to an existing integer became a bigint.

app/services/tool_filter_service.rb
```diff
@@ -0,0 +1,76 @@
...
+  def match_params(tool_group, params)
+    country = params.dig("filter", "country")&.upcase
+    languages = params.dig("filter", "language") || []
+    openness = params.dig("filter", "openness")&.to_i
+    confidence = params.dig("filter", "confidence")&.to_i
+
+    # Country Rules
+    if tool_group.rule_countries.any? { |o| o.negative_rule && o.countries.include?(country) }
+      return false
+    elsif tool_group.rule_countries.any? { |o| !o.negative_rule && o.countries.exclude?(country) }
```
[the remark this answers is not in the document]
cesarediaz: Good idea! I'll modify it

db/schema.rb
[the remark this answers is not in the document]
cesarediaz: > You were right: Rails defaults to bigint for new primary keys since 5.1: [rails/rails#26266](https://github.com/rails/rails/pull/26266)
> 
> That doesn't explain, however, why this reference to an existing integer became a bigint.
@knutsenm 
I think this is because when core team though about this change, it was with the intention to implement from there to the future the bigint for a reference no matter if this was represented as integer cuz they think that is better to cover this possibility to migrate later just the reference origin type and not the entire relationship, I mean, just adjust the origin id in table and not all the things.. origin and reference value.

I think the main idea is to prevent future big work in migrations, against a low cost of performance now cuz the databases are addapting each time better to work with big numbers without a big impact in performance

spec/acceptance/resources_controller_spec.rb
cesarediaz: @andrewroth @knutsenm 

I made changes here, when you ready take a look if this looks better now.

andrewroth:
Looks good! Can you get coverage to 100%? There's only a few lines left.

cesarediaz:
> Looks good! Can you get coverage to 100%? There's only a few lines left.

Ok, I think the goal was accomplished. All checks in green now

andrewroth (approved):
Looks good!

PR #1349, opened 2023-08-08
[GT-2017] : Provide a flag on auth requests to indicate if a user should be created or not

cesarediaz:
**When user Exists**

- [x] if create_user = true  should  => return error  user already exists
- [x] if create_user = false should  =>  return the user
- [x] if create_user = nil should => return the user

**When user Do Not Exists**

- [x] if create_user = true  should  => return  the user
- [x] if create_user = false should  =>  return error user does not exists
- [x] if create_user = nil should => return the user


app/controllers/auth_controller.rb
```diff
@@ -4,30 +4,40 @@ class AuthController < ApplicationController
...
+    message = []
     method = THIRD_PARTY_AUTH_METHODS.detect { |method|
       data_attrs[:"#{method}_access_token"] || data_attrs[:"#{method}_id_token"] ||
         data_attrs["#{method}_code"] || data_attrs["#{method}_token"]
     }
     token = case method
     when :apple_auth
       # special case for apple, which has given and family name passed in
-      user, apple_refresh_token = AppleAuthService.find_user_by_auth_code(data_attrs[:apple_auth_code], data_attrs[:apple_given_name], data_attrs[:apple_family_name])
+      user, message, apple_refresh_token = AppleAuthService.find_user_by_auth_code(data_attrs[:apple_auth_code], data_attrs[:apple_given_name], data_attrs[:apple_family_name], data_attrs[:create_user])
```
frett, line 15: maybe instead of returning an optional error message from the various auth functions, instead raise a custom error `UserNotFound` or `UserAlreadyExists` and then rescue that error to produce the error responses

cesarediaz: ok, is a good idea to implement

cesarediaz: implemented custom errors

frett:
I'm getting a `user_not_found` error when I'm authenticating without sending along the `create_user` flag. This should continue to function how it previously functioned and create the user if they don't exist.

frett:
When I send along `create_user=false` when I already have an account, it creates a brand new user for me

frett:
both of these tests were performed using facebook auth

cesarediaz:
> both of these tests were performed using facebook auth

can you send me a curl example to try with Postman?

frett:
@andrewroth this is ready for review

spec/acceptance/auth_controller_spec.rb
```diff
@@ -287,26 +400,69 @@
...
+            expect(response["errors"][0]["code"]).to eq("user_not_found")
+            expect(response["errors"][0]["detail"]).to eq("User account not found.")
+          end
 
-          expect(status).to be(201)
-          data = JSON.parse(response_body)["data"]
-          expect(data["attributes"]["user-id"]).to eq(user.id)
-          expect(data["attributes"]["token"]).to match(jwt_regex)
-          expect(data["attributes"]["apple-refresh-token"]).to eq(verify_auth_code_response["refresh_token"])
+          it "passing flag 'create_user: true' it succedess" do
```
andrewroth, line 415: `succedess` typo

app/concerns/common_service_auth_user_methods.rb
```diff
@@ -0,0 +1,33 @@
+# app/concerns/common_service_auth_methods.rb
+module CommonServiceAuthUserMethods
```
andrewroth, line 2: I don't agree with this being in concerns. We have a services folder already, why not put it in there? Better yet, eliminate the service altogether and put it right in `base_auth_service.rb` since it's only called from its subclasses. What do you think @knutsenm ?

andrewroth (changes requested):

app/concerns/common_service_auth_user_methods.rb
frett: I think part of the reason this got moved out to a separate chunk of code is that it's not only used by subclasses because the Okta auth doesn't currently subclass `base_atuh_service.rb`.

I'm not opposed to Okta subclassing the `base_auth_service.rb`, I'm guessing it just got missed because Okta was implemented a year or more before the other login methods.

cesarediaz:
> I'm getting a `user_not_found` error when I'm authenticating without sending along the `create_user` flag. This should continue to function how it previously functioned and create the user if they don't exist.

@frett I think it was solved

[not-new-user-when-not-send-flag.webm](https://github.com/CruGlobal/mobile-content-api/assets/51219/0e9f207e-4009-40e3-8799-8e344ccc26a0)

app/concerns/common_service_auth_user_methods.rb
andrewroth: @frett It does, no? https://github.com/CruGlobal/mobile-content-api/blob/master/app/services/okta_auth_service.rb#L3 has it extending `base_auth_service.rb`

If it's up to me, I would move it into BaseAuthService, and change the methods from class methods to regular instance methods.

frett: hmm, it does. I thought there was one that didn't extend it when I looked at it a couple weeks ago

frett: @cesarediaz since all the auth services extend it, we should probabyl put that logic in the `base_auth_service.rb`

cesarediaz: ok @frett

cesarediaz: @frett @andrewroth I have moved all logic and specs to BaseAuthService side

frett:
@andrewroth I did some cleanup to reduce code duplication and get specs to 100%, plus I believe Cesar addressed all your other outstanding comments.

Once you take a final look we can merge this

app/services/base_auth_service.rb
```diff
@@ -6,13 +6,42 @@ class BaseAuthService
...
   class << self
-    def find_user_by_token(access_token)
+    def user_existence_validation(create_user, users)
+      raise ::UserAlreadyExist::Error if user_already_exist(create_user, users)
+      raise ::UserNotFound::Error if user_not_found(create_user, users)
+    end
+
+    def first_or_initialize_user(primary_key, id, user_atts)
+      user = User.where(primary_key => id).first_or_initialize
+      user.update!(user_atts)
```
andrewroth, line 16: `first_or_initialize_user` is not really an accurate name because of this `update!` call. `first_or_create_user` would be more accurate.

app/services/base_auth_service.rb
```diff
@@ -25,10 +54,18 @@ def validate_expected_fields!(decoded_token)
...
-      user.update!(user_atts)
-      user
+    def setup_user(remote_user_id, user_atts, create_user)
+      users = User.where(primary_key => remote_user_id)
+
+      user_existence_validation(create_user, users)
+      user = existent_user(create_user, users)
+      return user if user
+
+      if create_user.nil?
```
andrewroth, line 64: create_user is a boolean, right? So technically, create_user could be false, and false.nil? will evaluate to false. Why not put the if structure with the positive case first, so `if create_user`... 

I'm also kind of confused by this just in general. So if you *don't* want to create the user, first_or_initialize_user will still create a new record.... ?

I think having these helper methods (first_or_create_user and new_user) makes reading it more complex; my gut feeling is these could be done inline here and it would avoid having to mentally jump to another method to follow the code.

andrewroth: Oh, I see! A comment to that effect would probably be sufficient then.

frett: @andrewroth I reworked the logic

andrewroth (approved):

PR #1365, opened 2023-08-21
[GT-1563] API - Create a user specific key-value store API

cesarediaz:


config/routes.rb
```diff
@@ -58,6 +58,7 @@
   patch "users/:user_id/counters/:id", to: "user_counters#update"
   get "users/:id", to: "users#show"
   delete "users/:id", to: "users#destroy"
+  patch "users/me", to: "users#me"
```
frett, line 61: this should be `users/:id`

`users/me` is a shortcut that resolves to the current users id

app/controllers/users_controller.rb
```diff
@@ -10,6 +10,12 @@ def destroy
     render json: "", status: 204
   end
 
+  def me
```
frett, line 13: this method should not be named `me`, the isn't relevant to what this method does at all. it should probably be named `update`

app/serializers/user_serializer.rb
```diff
@@ -4,6 +4,7 @@ class UserSerializer < ActiveModel::Serializer
   type "user"
   attributes :sso_guid, :created_at, :given_name, :family_name, :name
   has_many :tools, key: "favorite-tools", serializer: ResourceFavoritedSerializer
+  has_many :user_attributes
```
frett, line 7: we don't need to expose the raw `user_attributes` relationship through the serializer, this relationship is an internal implementation detail and nothing that we care about from API clients

app/serializers/user_attribute_serializer.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class UserSerializer < ActiveModel::Serializer
+  belongs_to :user
+end
```
frett, line 5: there should be no need to define a `user_attribute_serializer`

cesarediaz:
@frett I think that all  previous comments have been addressed. Please take a look when you ready

spec/acceptance/users_controller_spec.rb
```diff
@@ -11,7 +11,7 @@
   let(:resource) { Resource.first }
   let(:resource2) { Resource.second }
 
-  patch "users/me" do
+  patch "users/update" do
```
frett, line 14: this acceptance test can still run against `users/me`, it should automatically resolve to the current user

app/controllers/users_controller.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class UsersController < WithUserController
+  before_action :get_user, only: [:update]
+
```
frett, line 5: this before_action is unnecessary, `before_action :authorize!` defined in `with_user_controller` will correctly populate the `@user` model for you

cesarediaz: done

frett:
When deleting a user I encounter the following FK violation:

Screenshot 2023-08-22 at 4 39 00 PM

We need to make sure deleting a user will cascade and delete any key-value attributes set on that user

cesarediaz:
@frett I've solved I think the `PG::ForeignKeyViolation: ERROR:` . Take a look when you ready

app/models/user.rb
```diff
@@ -3,6 +3,24 @@ class User < ApplicationRecord
...
   validates :email, presence: true
+
+  def set_arbitrary_attributes!(data_attrs)
+    data_attrs.each_pair do |key, value|
+      attr_name = key[/^attr-(.*)$/, 1]&.downcase
+      next unless attr_name
+      attr_name.tr!("-", "_")
+      attribute = user_attributes.where(key: attr_name).first_or_initialize
+
+      if value.nil?
```
andrewroth, line 18: I think this might flow nicer to put the positive case first, `if value`, `else`

andrewroth (changes requested):
Looks good, just the if flow thing, which is pretty minor

app/models/user.rb
cesarediaz: done

andrewroth (approved):

PR #1370, opened 2023-08-25
GT-1962 Publish translations via sidekiq

frett:
I'm creating this PR on behalf of @cesarediaz


frett:
@knutsenm @andrewroth @jbirdjavi I'm guessing the sidekiq config is currently incomplete and not setup to use redis or run the tasks on our separate ECS container. can you give feedback to help out @cesarediaz with what needs to change?

app/controllers/resources_controller.rb
```diff
@@ -38,8 +38,44 @@ def suggestions
     render json: ToolFilterService.new(params).call, include: params[:include], fields: field_params, status: :ok
   end
 
+  def publish_translation
+    if valid_publish_params?
+      publish_translations
+      head :no_content
```
frett, line 44: @cesarediaz from the sample request & response I had provided this method should return the `Translation` objects that were created.

config/routes.rb
```diff
@@ -86,6 +89,7 @@
     end
   end
 
+  mount Sidekiq::Web => "/sidekiq"
```
frett, line 92: @knutsenm @andrewroth I'm assuming we don't expose a sidekiq route like this?

app/controllers/resources_controller.rb
```diff
@@ -38,8 +38,44 @@ def suggestions
...
+    end
+  end
+
   private
 
+  def valid_publish_params?
+    params["data"] &&
+      params["data"]["relationships"] &&
+      params["data"]["relationships"]["languages"] &&
+      params["data"]["relationships"]["languages"]["data"] &&
```
andrewroth, line 56: I think you could use `dig` here to make this more succinct.

config/routes.rb
andrewroth: Usually there's an admin check around it..

.github/workflows/ruby.yml
```diff
@@ -34,6 +34,8 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           bundler-cache: true
+        env:
+          BUNDLE_GEMS__CONTRIBSYS__COM: ${{ secrets.BUNDLE_GEMS__CONTRIBSYS__COM }}
```
knutsenm, line 38: The same step in `lint:` needs this as well.

Also, you need to configure Redis in the `services:` section above:

```
      redis:
        image: redis
        options: --health-cmd "redis-cli ping" --health-interval 10s --health-timeout 5s --health-retries 5
        ports:
          - 6379:6379
```

app/sidekiq/publish_translation_job.rb
```diff
@@ -0,0 +1,9 @@
+class PublishTranslationJob
+  include Sidekiq::Job
+  queue_as :default
```
knutsenm, line 3: The Sidekiq syntax we tend to use is `sidekiq_options queue: ...`, but it's not necessary to take the default.

config/application.rb
```diff
@@ -33,5 +33,7 @@ class Application < Rails::Application
 
     ActiveModelSerializers.config.adapter = :json_api
     FileUtils.mkdir_p("pages")
+
+    config.active_job.queue_adapter = :sidekiq
```
knutsenm, line 37: This is not necessary if you're not using ActiveJob, which, due to `include Sidekiq::Job`, you are not.

config/routes.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "sidekiq/web"
```
knutsenm, line 3:
```suggestion
require "sidekiq/pro/web"
```

config/routes.rb
knutsenm: Yes; Global Registry, for example, uses Basic Auth:
https://github.com/CruGlobal/global_registry/blob/master/config/initializers/sidekiq.rb#L44

knutsenm (changes requested):

Gemfile
```diff
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 source "https://rubygems.org"
+source 'https://gems.contribsys.com/' do
+  gem 'sidekiq-pro', '< 6'
+end
```
knutsenm, line 6: @cesarediaz for local development only, you will want to adjust this:
```suggestion
gem "sidekiq", "~> 6.0"
```

Also, I suggest you follow standard and use double quotes.

config/routes.rb
knutsenm: ...but for local development only, omit the `/pro`

app/controllers/resources_controller.rb
cesarediaz: done

config/routes.rb
cesarediaz: @knutsenm I do not have access to the line of code you mention there... how do you say to implement it? is the next the idea you expressed plus some basic auth somewhere in controllers?

```ruby
  namespace :admin do
    mount Sidekiq::Web => "/sidekiq"
  end
```

knutsenm: No; by "admin check" he meant authorization against some secret, like an API key for instance.

https://github.com/sidekiq/sidekiq/wiki/Monitoring#authentication has some examples.

config/application.rb
cesarediaz: done

config/routes.rb
cesarediaz: done

app/sidekiq/publish_translation_job.rb
cesarediaz: @knutsenm we can implement then like this next

```ruby
  sidekiq_options queue: :publish_translation
```

what do you think?

config/routes.rb
cesarediaz: done

app/sidekiq/publish_translation_job.rb
knutsenm: There will likely not be a need for multiple queues, so the default should suffice; @frett ?

config/routes.rb
```diff
@@ -86,6 +89,9 @@
     end
   end
 
+  authenticate :user, lambda { |u| u.admin? } do
```
knutsenm, line 92: > undefined method `authenticate' for #<ActionDispatch::Routing::Mapper

This is not valid syntax, You are not using Devise.

Gemfile
cesarediaz: @knutsenm  I'm working in localhost with  the gem like 
```ruby
  group :development, :test do
    gem "sidekiq", "~> 7.1", ">= 7.1.2"
  end
```
but when pushing code I ommit it

this is what do you mean... avoid to work with a "pro" sidekiq gem on my localhost?

cesarediaz: now I just will fix the single quote by replacing them for double ones on the sidekiq pro lines until you answer me

knutsenm: > this is what do you mean... avoid to work with a "pro" sidekiq gem on my localhost?

Yes, but stick with sidekiq 6

config/environments/staging.rb
```diff
@@ -10,4 +10,6 @@
 Rails.application.configure do
   config.secret_key_base = ENV["SECRET_KEY_BASE"]
   config.action_cable.disable_request_forgery_protection = true
+
+  config.active_job.queue_adapter = :sidekiq
```
knutsenm, line 14: Remove this, too.

knutsenm (changes requested):

app/controllers/resources_controller.rb
```diff
@@ -38,8 +38,42 @@ def suggestions
...
+    languages = params["data"]["relationships"]["languages"]["data"]
+
+    languages.each do |lang|
+      translations << publish_translation_for_language(lang)
+    end
+    translations
+  end
+
+  def publish_translation_for_language(language_data)
+    translation = find_latest_translation(language_data["id"])
```
frett, line 66: this should be find or create a draft translation. right now the scope you are using is just finding the most recent translation whether it's been published or not. In reality you should looking for a translation that has `is_published = false`, and if that doesn't exist create a new Translation in that state.

config/routes.rb
knutsenm: Whatever @frett says. It would be nice if you could leverage Okta login and the User#admin boolean you already have, but there is no Web front end so that may be more trouble than it's worth.

frett: I agree it would be a nice tie in, but I think the cost would greatly outweigh the benefit. Let's use basic auth that checks user & pass against a couple environment variables.

```
  Sidekiq::Web.use Rack::Auth::Basic do |username, password|
    username == ENV.fetch("SIDEKIQ_USERNAME") && password == ENV.fetch("SIDEKIQ_PASSWORD")
  end
```

cesarediaz: @knutsenm right.. ok

app/controllers/resources_controller.rb
cesarediaz: ok, understood!

app/controllers/resources_controller.rb
```diff
@@ -38,8 +38,45 @@ def suggestions
...
+
+  def publish_translation_for_language(language_data)
+    draft_translation = find_or_create_latest_translation(language_data["id"])
+    if draft_translation
+      PublishTranslationJob.perform_async(draft_translation.id)
+      draft_translation
+    end
+  end
+
+  def find_or_create_latest_translation(language_id)
```
frett, line 73: let's name the method `find_or_create_draft_translation` to make it more clear that it's working with a "draft" translation

app/controllers/resources_controller.rb
```diff
@@ -38,8 +38,45 @@ def suggestions
...
+    resource = Resource.find(params["resource_id"])
+
+    draft = resource.create_draft(language_id) if resource
```
frett, line 76: it looks like the `create_draft` method will always create a new draft even if another draft already exists for that resource & language.

frett: We want to re-use an existing draft when it exists, so that if there is an error publishing the draft we don't just keep creating new draft translations.

cesarediaz: ok, I'll revise it

app/models/translation.rb
```diff
@@ -76,6 +81,20 @@ def manifest_translated_phrases
...
+      ActiveRecord::Base.transaction do
+        phrases = manifest_translated_phrases
+        name_desc_onesky(phrases)
+        create_translated_attributes(phrases)
+      end
+    end
+
+    Package.new(self).push_to_s3
+  rescue => e
+    Rollbar.error(e)
```
frett, line 95: as far as I know, from the admin tool we will have no way of accessing what this error is or associating it with a specific translation.

Can we store the error string on a new field on the translation object (`publishing-error`) that we will be able to read from the admin tool out of the translation draft model?

cesarediaz: @frett sorry, I had changed this last night but I forgot to commit :facepalm:

app/sidekiq/publish_translation_job.rb
```diff
@@ -0,0 +1,13 @@
+class PublishTranslationJob
+  include Sidekiq::Job
+
+  def perform(id)
+    translation = Translation.find(id)
+    begin
+      translation.push_published_to_s3
+      translation.update!(is_published: true)
+    rescue => e
+      translation.errors.add(:error, e)
```
frett, line 10: hmm, are these `errors` stored in the database to where they can be present in the serializer on subsequent requests?

If not I'm fine with this being a simple text field on the `Translation` model that tracks publishing errors

app/models/translation.rb
```diff
@@ -76,6 +81,18 @@ def manifest_translated_phrases
     @manifest_translated_phrases ||= download_translated_phrases("name_description.xml")
   end
 
+  def push_published_to_s3
```
frett, line 84: We should sort-circuit publishing if the translation has already been published

```suggestion
  def push_published_to_s3
    return if is_published
```

app/sidekiq/publish_translation_job.rb
```diff
@@ -0,0 +1,13 @@
+class PublishTranslationJob
+  include Sidekiq::Job
+
+  def perform(id)
+    translation = Translation.find(id)
+    begin
+      translation.push_published_to_s3
+      translation.update!(is_published: true)
```
frett, line 8: I think marking `is_published: true` should be part of the `push_published_to_s3`

app/models/translation.rb
```diff
@@ -15,9 +15,14 @@ class Translation < ActiveRecord::Base
   validates_with UsesOneskyValidator
 
   before_destroy :prevent_destroy_published, if: :is_published
-  before_update :push_published_to_s3
```
frett, line 18: I believe the legacy publishing process triggered this update from the `drafts_controller`, if that is the case we should update that logic to trigger `push_published_to_s3` from there if the request indicates `is_published: true`

app/controllers/drafts_controller.rb
```diff
@@ -47,7 +47,13 @@ def language_id
...
   def edit
     translation = load_translation
-    translation.update_draft(data_attrs)
+
+    # Note that this endpoint isn't used for anything other than publishing to s3. Eventually once the new client switches
+    # to the new way of publishing, we'll delete this file
+    # see: https://jira.cru.org/browse/GT-2182
+    do_publishing = data_attrs[:is_published]
+
+    translation.push_published_to_s3
```
frett, line 56: I don't think we ever call this endpoint with anything other than `is_published: true`, but to be safe we probably should protect the call to `translation.push_published_to_s3` with `if do_publishing`

app/controllers/resources_controller.rb
```diff
@@ -38,8 +38,45 @@ def suggestions
...
+      draft_translations << publish_translation_for_language(lang)
+    end
+    draft_translations
+  end
+
+  def publish_translation_for_language(language_data)
+    draft_translation = find_or_create_draft_translation(language_data["id"])
+    if draft_translation
+      PublishTranslationJob.perform_async(draft_translation.id)
+      draft_translation
```
frett, line 69: before queuing up the PublishTranslationJob let's clear any previous `publishing_errors`. That way the admin isn't shown outdated errors

app/models/translation.rb
```diff
@@ -76,6 +77,21 @@ def manifest_translated_phrases
...
+    if resource.uses_onesky?
+      ActiveRecord::Base.transaction do
+        phrases = manifest_translated_phrases
+        name_desc_onesky(phrases)
+        create_translated_attributes(phrases)
+      end
+    end
+
+    Package.new(self).push_to_s3
+    update(is_published: true)
```
frett, line 92: let's also clear any `publishing_errors` here since we have just finished successfully publishing the translation.

spec/models/translation_spec.rb
```diff
@@ -162,9 +162,7 @@
     it "uploads the translation to S3" do
       package = double
       allow(Package).to receive(:new).and_return(package)
-      allow(package).to receive(:push_to_s3)
-
-      translation.update(is_published: true)
+      allow(package).to receive(:push_published_to_s3)
```
frett, line 165: I believe this should still be `:push_to_s3` since it's on the Package class and not the Translation class

spec/package_spec.rb
```diff
@@ -372,16 +372,6 @@ def pages_dir_nil
...
 
-  def mock_onesky(project_id = nil)
-    ENV["ONESKY_API_SECRET"] ||= ""
-    project_id ||= Resource.find(1).onesky_project_id
-    response = RestClient::Response.new('{ "1":"value" }')
-    response.instance_variable_set :@code, 200
-    allow(RestClient).to receive(:get)
-      .with("https://platform.api.onesky.io/1/projects/#{project_id}/translations", any_args)
-      .and_return(response)
-  end
```
frett, line 383: I'm curious about this mock being removed, none of this work should have touched OneSky interactions (other than them being transitively triggered within the publish process which now fires from sidekiq

frett: nm, I see it was moved to a different location

app/controllers/resources_controller.rb
```diff
@@ -38,8 +38,46 @@ def suggestions
     render json: ToolFilterService.new(params).call, include: params[:include], fields: field_params, status: :ok
   end
 
+  def publish_translation
```
frett, line 41: This is not behind authorization currently, it should be added to the `:authorize` `before_action`

frett:
sidekiq isn't processing queued tasks. It looks like the sidekiq ECS task is failing to start

Screenshot 2023-11-06 at 1 41 21 PM

config/routes.rb
```diff
@@ -87,6 +90,13 @@
     end
   end
 
+  if Rails.env.production?
+    Sidekiq::Web.use Rack::Auth::Basic do |username, password|
+      username == ENV.fetch("SIDEKIQ_USERNAME") && password == ENV.fetch("SIDEKIQ_PASSWORD")
+    end
+  end
+  mount Sidekiq::Web, at: "/sidekiq"
```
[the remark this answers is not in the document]
frett: maybe because this is checking for the production environment, this should also apply to our staging environment since that is hosted publicly

app/sidekiq/publish_translation_job.rb
```diff
@@ -0,0 +1,12 @@
+class PublishTranslationJob
+  include Sidekiq::Job
+
+  def perform(id)
+    translation = Translation.find(id)
+    begin
+      translation.push_published_to_s3
+    rescue => e
+      translation.update(publishing_errors: e.to_s)
```
frett, line 9: `publishing_errors` is currently missing from the translation serializer

app/sidekiq/publish_translation_job.rb
```diff
@@ -0,0 +1,12 @@
+class PublishTranslationJob
```
frett, line 1: it doesn't look like this job is correctly following the unique job constraint. I was able to quickly queue up 5 of the same request. I might just be misunderstanding how the unique job constraint works/is displayed in the sidekiq console

Screenshot 2023-11-07 at 1 40 28 PM

frett: This has been fixed

frett:
This looks good from a functional standpoint, just needs code review from @knutsenm and then we're good to merge it

app/models/translation.rb
```diff
@@ -76,6 +71,21 @@ def manifest_translated_phrases
     @manifest_translated_phrases ||= download_translated_phrases("name_description.xml")
   end
 
+  def push_published_to_s3
+    return if is_published
```
knutsenm, line 75: Did the meaning of that boolean reverse? The old routine had `return unless`.

config/routes.rb
```diff
@@ -87,6 +90,13 @@
     end
   end
 
+  if Rails.env.production? || Rails.env.staging?
+    Sidekiq::Web.use Rack::Auth::Basic do |username, password|
+      username == ENV.fetch("SIDEKIQ_USERNAME") && password == ENV.fetch("SIDEKIQ_PASSWORD")
+    end
+  end
+  mount Sidekiq::Web, at: "/sidekiq"
```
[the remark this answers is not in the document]
frett: it does prompt me for auth when trying to access https://mobile-content-api-stage.cru.org/sidekiq

app/sidekiq/publish_translation_job.rb
```diff
@@ -0,0 +1,14 @@
...
+  include Sidekiq::Job
+
+  sidekiq_options lock: :until_executed
+
+  def perform(id)
+    translation = Translation.find(id)
+    begin
+      translation.push_published_to_s3
+    rescue => e
+      translation.update(publishing_errors: e.to_s)
```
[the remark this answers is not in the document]
frett: yes, the purpose of `publishing_errors` is to capture any publishing failures to expose to the content admin UI and allow the content author to address them and retry publishing again afterwards.

We actually clear previous errors when we queue up the draft translation to be published and also when the translation is successfully published.

app/models/translation.rb
frett: previously publishing was a side effect of changing the `is_published` flag to true using an `before_update` hook.

We have refactored the publishing logic to be directly triggered and once it succeeds we update the `is_published` flag to true

frett: We have always had the constraint that a specific translation can only be published once and is considered immutable after it is successfully published

spec/models/resource_spec.rb
```diff
@@ -28,7 +28,7 @@
     end
 
     context "existing resource/language combination" do
-      it "adds a new version" do
+      xit "adds a new version" do
```
knutsenm, line 31: Do you mean for this to remain disabled? If so, perhaps remove the test?

(Same question, retroactively, about spec/services/okta_auth_service_spec.rb:49)

knutsenm (approved):
💯

PR #1373, opened 2023-08-28
[GT-1852] API support (the) tracking (of) training tips (that) a user has completed

cesarediaz:
**Training Tips**

- [x] add route
- [x] add migration
- [x] add model && controller
- [x] add serializer
- [x] add specs


app/models/training_tip.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class TrainingTip < ActiveRecord::Base
```
frett, line 3: I'm not seeing a relationship back to a user from this model.

config/routes.rb
```diff
@@ -37,6 +37,8 @@
     delete "tools/:id", to: "tool_groups#delete_tool"
   end
 
+  resources :training_tips, path: "training-tips"
```
frett, line 40: this route should be under a specific user, so `/users/:user_id/training-tips`

app/controllers/training_tips_controller.rb
```diff
@@ -0,0 +1,18 @@
+class TrainingTipsController < ApplicationController
+  before_action :authorize!
```
frett, line 2: this needs to also be checking user authorization, so you should extend the `WithUserController` instead to provide that functionality for you

app/models/training_tip.rb
cesarediaz: ok, I will add this relationship there

app/controllers/training_tips_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def create
+    create_training_tip
+  end
+
+  private
+
+  def create_training_tip
+    created = TrainingTip.create!(permit_params(:tool, :locale, :tip_id, :is_completed))
+    response.headers["Location"] = "tool_groups/#{created.id}"
```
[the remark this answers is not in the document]
cesarediaz: done

app/controllers/training_tips_controller.rb
cesarediaz: done

app/models/training_tip.rb
cesarediaz: done

config/routes.rb
```diff
@@ -67,6 +65,10 @@
   end
   delete "users/me/relationships/favorite-tools", to: "favorite_tools#destroy"
 
+  scope "users/me" do
```
frett, line 71: I forgot to mention that this should leverage the pre-existing `uesrs/:user_id` logic for the route. that logic will resolve `user_id=me` to be the current user and in the future support admin management functionality for managing other users if necessary.

This change should also apply to the `relationships/favorite-tools` endpoint as well

db/migrate/20230825233539_create_user_training_tip.rb
```diff
@@ -0,0 +1,15 @@
...
+      t.references :user, null: false, foreign_key: true
+      t.references :tool, null: false, foreign_key: {to_table: :resources}
+      t.references :language, null: false, foreign_key: true
+      t.string :tip_id
+      t.boolean :is_completed
+
+      t.timestamps
+    end
+
+    add_index :user_training_tips, [:id, :tool_id, :language_id, :tip_id], unique: true, name: "training-tips-unique-index"
```
frett, line 13: I don't think you want `:id` in this unique index

frett: unless `:id` is the reference to the `:user` relationship which isn't currently present in this index

andrewroth: Oh, yeah, it should be `user_id`

app/controllers/training_tips_controller.rb
```diff
@@ -0,0 +1,32 @@
+class TrainingTipsController < WithUserController
+  before_action :convert_hyphen_to_dash, only: [:create]
+
+  def create
+    user_training_tip = @user.user_training_tips.create!(permitted_params)
+    response.headers["Location"] = "users/me/training-tips/#{user_training_tip.id}"
```
frett, line 6: the `Location` header should use `user_id` for accuracy.

app/controllers/training_tips_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+    render json: user_training_tip, status: :created
+  rescue ActiveRecord::RecordInvalid => e
+    render json: {errors: formatted_errors("record_invalid", e)}, status: :unprocessable_entity
+  end
+
+  def update
+    user_training_tip = @user.user_training_tips.find(params[:id])
+    user_training_tip.update(permitted_params)
```
knutsenm, line 14: Most (all?) of the other controllers do `update!`

...and `destroy!`, below.

spec/factories/training_tips.rb
```diff
@@ -0,0 +1,9 @@
+FactoryBot.define do
+  factory :training_tip do
```
knutsenm, line 2: No such model; you can delete this factory I think.

knutsenm (changes requested):

knutsenm (approved):

frett (approved):

PR #1395, opened 2023-09-20
GT-1562 Support additional fields for favorite-tools API

frett:


app/controllers/concerns/fields.rb
```diff
@@ -1,8 +1,8 @@
 module Fields
   private
 
-  def field_params
-    return {} unless params[:fields]
+  def field_params(default = {})
+    return default unless params[:fields]
```
frett, line 5: long term we probably should preserve default keys that aren't specified as a `fields[{key}]` parameter in the request.

But for now just being able to override the default fields is enough for the current use-case.

app/serializers/resource_serializer.rb
```diff
@@ -19,9 +19,13 @@ class ResourceSerializer < ActiveModel::Serializer
...
   belongs_to :default_variant, key: "default-variant", if: -> { object&.resource_type&.name == "metatool" }
 
-  def attributes(*args)
+  def attributes(requested_attrs = nil, reload = false)
     hash = super
-    object.resource_attributes.each { |attribute| hash["attr_#{attribute.key}"] = attribute.value }
+    object.resource_attributes.each { |attribute|
+      key = "attr-#{attribute.key}".tr("_", "-").to_sym
+      next unless requested_attrs.nil? || requested_attrs.include?(key)
+      hash[key] = attribute.value
```
frett, line 27: the serializer does some bizarre `_` to `-` and back translations. This is probably due to the default syntax for jsonapi being to use kebab case vs ruby/rails preferring snake case.

PR #1401, opened 2023-09-26
GT-2143 Fix fields parameter for resource serializer

frett:
- filter arbitrary resource_attributes based upon the fields parameter
- make the resource serializer keys symbols to correctly handle them with the fields parameter

### DIFF
--- a/app/serializers/resource_serializer.rb
+++ b/app/serializers/resource_serializer.rb
@@ -19,9 +19,13 @@ class ResourceSerializer < ActiveModel::Serializer
 
   belongs_to :default_variant, key: "default-variant", if: -> { object&.resource_type&.name == "metatool" }
 
-  def attributes(*args)
+  def attributes(requested_attrs = nil, reload = false)
     hash = super
-    object.resource_attributes.each { |attribute| hash["attr_#{attribute.key}"] = attribute.value }
+    object.resource_attributes.each { |attribute|
+      key = "attr-#{attribute.key}".tr("_", "-").to_sym
+      next unless requested_attrs.nil? || requested_attrs.include?(key)
+      hash[key] = attribute.value
+    }
     hash
   end
 end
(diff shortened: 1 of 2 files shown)


app/serializers/resource_serializer.rb
```diff
@@ -2,8 +2,9 @@
 
 class ResourceSerializer < ActiveModel::Serializer
   type "resource"
-  attributes :id, :name, :abbreviation, :description, :onesky_project_id, :total_views, :manifest
-  attribute :resource_type_name, key: "resource-type"
+  attributes :id, :name, :abbreviation, :description, :onesky_project_id, :manifest
+  attribute :resource_type_name, key: :"resource-type"
+  attribute :total_views, key: :"total-views"
```
jbirdjavi, line 7: I'm really surprised this works since I've never seen a symbol with quotes like that, but it does work so that's great.

app/serializers/resource_serializer.rb
```diff
@@ -19,9 +20,13 @@ class ResourceSerializer < ActiveModel::Serializer
 
   belongs_to :default_variant, key: "default-variant", if: -> { object&.resource_type&.name == "metatool" }
 
-  def attributes(*args)
+  def attributes(requested_attrs = nil, reload = false)
```
jbirdjavi, line 23: The `reload` parameter doesn't appear to be used here, and I'm not sure where it's coming from. Can you shed some light on that?

frett: it's on the overloaded method, if I leave it off I believe it was throwing some errors

frett: I take that back, it's not throwing errors, it must have been elsewhere I was seeing the errors

app/serializers/resource_serializer.rb
jbirdjavi: Nope, you can leave it. It's new to me, so I'm glad that I know it works. I might have to play around with it to see how it works in other situations.

jbirdjavi (approved):

PR #1459, opened 2023-12-08
Trial at making Suggestions endpoint return results for negative rules

dr-bizz:
## Description
The `/resources/suggestions` API currently doesn't allow me to filter on negative rules. From reading the PR which Cezar did, I've tried to add this functionality. I'm unsure if it will work (prob not), but I wanted to give it a shot and maybe learn something.

## Changes
- I've allowed for 3 new params `country_negative_rule`, `language_negative_rule`, `praxis_negative_rule`.
- I've updated the IF statements to use these rules.
- If the negative rule is `nil` or `false`, it should act as normal. But if true, it should ensure we exclude the filtered items.

### DIFF
--- a/app/services/tool_filter_service.rb
+++ b/app/services/tool_filter_service.rb
@@ -45,29 +45,34 @@ class ToolFilterService
     languages = params.dig("filter", "language") || []
     openness = params.dig("filter", "openness")&.to_i
     confidence = params.dig("filter", "confidence")&.to_i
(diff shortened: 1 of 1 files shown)


frett:
having negative rule booleans doesn't make sense for this API. If a negative rule is defined for a tool group, then that rule will only match if the value sent to the suggestions API isn't specified in the rule.

for example if there is a negative language rule containing French, it will only match on the suggestions API if the language specified is not french.

PR #1544, opened 2024-03-13
GT-2289 Better define how multiple events and listeners interact with each other

frett:
This PR better defines how events and listeners interact with each other and explicitly states when various listeners are "active"

### DIFF
--- a/public/xmlns/content.xsd
+++ b/public/xmlns/content.xsd
@@ -435,7 +435,9 @@
                         </xs:sequence>
                         <xs:attribute name="listeners" type="content:listeners">
                             <xs:annotation>
-                                <xs:documentation>event_ids that trigger this tab</xs:documentation>
+                                <xs:documentation>event_ids that trigger this tab. These listeners are only "active"
+                                    when the containing tabs content element is in the current active content stack.
+                                </xs:documentation>
                             </xs:annotation>
                         </xs:attribute>
                     </xs:complexType>
--- a/public/xmlns/content.xsd
+++ b/public/xmlns/content.xsd
@@ -722,12 +722,16 @@
             <xs:attribute name="loop" type="xs:boolean" default="true" />
             <xs:attribute name="play-listeners" type="content:listeners">
                 <xs:annotation>
-                    <xs:documentation>event_ids that trigger playback of this animation.</xs:documentation>
+                    <xs:documentation>event_ids that trigger playback of this animation. These listeners are only
+                        "active" when the animation is not already playing and is in the current active content stack.
+                    </xs:documentation>
                 </xs:annotation>
             </xs:attribute>
             <xs:attribute name="stop-listeners" type="content:listeners">
                 <xs:annotation>
-                    <xs:documentation>event_ids that will stop playback of this animation.</xs:documentation>
+                    <xs:documentation>event_ids that will stop playback of this animation. These listeners are only
+                        "active" when the animation is already playing and is in the current active content stack.
+                    </xs:documentation>
                 </xs:annotation>
             </xs:attribute>
 
--- a/public/xmlns/manifest.xsd
+++ b/public/xmlns/manifest.xsd
@@ -18,6 +18,17 @@
 
     <xs:annotation>
         <xs:documentation>
+            Content Stacks
+            --------------
+
+            All of the content within the content spec is built around the concept of a content stack. A content stack
+            is a vertical stack of content that supports vertical scrolling. We try to avoid doing absolute positioning
+            of content to avoid responsive issues with differing screen sizes.
+
+            Throughout the content spec we will refer to an active content stack to reference the current content stack
+            being rendered on the screen. For example for a tract tool showing a card, the active content stack refers
+            to all content elements that are in that card.
+
             Events and Listeners
             --------------------
 
--- a/public/xmlns/manifest.xsd
+++ b/public/xmlns/manifest.xsd
@@ -35,7 +35,8 @@
             Throughout the content spec there are elements that trigger events and listeners that listen for events.
             Whenever an element triggers an event or a series of events, each event triggered is processed in sequence.
             For each event any "active" listeners that match the element will fire. Once each triggered event is
-            processed, the next triggered event will be processed with the updated "active" listeners.
+            processed, the next triggered event will be processed with the updated "active" listeners. Unless otherwise
+            specified content visibility attributes have no impact on whether a listener is "active".
 
             For example, if tract page 1 is active with the hero active and an animation running in the hero. Then if a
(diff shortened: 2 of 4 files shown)


frett:
@aaronlaib @levieggertcru @caleballdrin @dr-bizz I'm assigning all of you to get feedback if the added documentation is clear around how events and listeners interact with each other.

Please leave comments if there is something you don't understand about how multiple events should be evaluated, when certain listeners can actually be triggered, or anything else related to events and listeners.

public/xmlns/content.xsd
```diff
@@ -722,12 +724,16 @@
...
+                    <xs:documentation>event_ids that trigger playback of this animation. These listeners are only
+                        "active" when the animation is not already playing and is in the current active content stack.
+                    </xs:documentation>
                 </xs:annotation>
             </xs:attribute>
             <xs:attribute name="stop-listeners" type="content:listeners">
                 <xs:annotation>
-                    <xs:documentation>event_ids that will stop playback of this animation.</xs:documentation>
+                    <xs:documentation>event_ids that will stop playback of this animation. These listeners are only
+                        "active" when the animation is already playing and is in the current active content stack.
```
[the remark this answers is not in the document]
frett: that's an interesting dynamic I hadn't considered, how would events interact with listeners on content that is hidden based upon state.

What I had meant with this statement (and I will try to come up with a more clear way of saying it) is content that is currently rendered, plus content you can scroll vertically to. But not content that takes switching to a different tool card or page.

So if there was multiple pages of content in a hero that you have to scroll to see it all, listeners for animations at the top or bottom of that content stack would still trigger, even if the animation is not currently scrolled onto the screen.

public/xmlns/manifest.xsd
```diff
@@ -16,6 +16,40 @@
...
+            Content Stacks
+            --------------
+
+            All of the content within the content spec is built around the concept of a content stack. A content stack
+            is a vertical stack of content that supports vertical scrolling. We try to avoid doing absolute positioning
+            of content to avoid responsive issues with differing screen sizes.
+
+            Throughout the content spec we will refer to an active content stack to reference the current content stack
+            being rendered on the screen. For example for a tract tool showing a card, the active content stack refers
+            to all content elements that are in that card.
```
frett, line 30: @levieggertcru I added this section to define what I meant by an active content stack

public/xmlns/manifest.xsd
```diff
@@ -16,6 +16,40 @@
...
+
+            Events and Listeners
+            --------------------
+
+            Throughout the content spec there are elements that trigger events and listeners that listen for events.
+            Whenever an element triggers an event or a series of events, each event triggered is processed in sequence.
+            For each event any "active" listeners that match the element will fire. Once each triggered event is
+            processed, the next triggered event will be processed with the updated "active" listeners. Unless otherwise
+            specified content visibility attributes have no impact on whether a listener is "active".
```
frett, line 39: I added a statement here that listeners aren't affected by content visibility attributes unless otherwise specified

public/xmlns/content.xsd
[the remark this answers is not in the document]
frett: I addressed this with a couple updates to the common documentation in manifest.xsd

public/xmlns/manifest.xsd
levieggertcru: Thanks @frett

levieggertcru (approved):

public/xmlns/content.xsd
```diff
@@ -435,7 +435,9 @@
                         </xs:sequence>
                         <xs:attribute name="listeners" type="content:listeners">
                             <xs:annotation>
-                                <xs:documentation>event_ids that trigger this tab</xs:documentation>
+                                <xs:documentation>event_ids that trigger this tab. These listeners are only "active"
```
dr-bizz, line 438: When you describe "active" do you mean, they are callable?

public/xmlns/page.xsd
```diff
@@ -32,10 +32,20 @@
...
-                <xs:documentation>event_ids that trigger this page</xs:documentation>
+                <xs:documentation>This is a list of event_ids that will cause navigation to jump to this page. Page
+                    listeners are always "active".
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+        <xs:attribute name="dismiss-listeners" type="content:listeners">
+            <xs:annotation>
+                <xs:documentation>This is the list of event_ids that will pop this page off a navigation stack. This
+                    listener is only "active" when the page is on top of the navigation stack.
```
dr-bizz, line 43: Is the navigation stack, the list of pages available?

dr-bizz (commented):
I added some comments, but overall I understand the documentation. - Thank you for writing this up.

public/xmlns/page.xsd
frett: the navigation stack is a concept for a non-linear tool type (cyoa). Each page is standalone and uses events/listeners to navigate to other pages. The dismiss listener is like a back button.

So if you start on a table of contents page, press button a which uses an event to trigger page a, then on page a there is a button that fires an event that matches the dismiss listener, it will dismiss page a and go back to the table of contents which is the next page down the page "stack"

public/xmlns/content.xsd
frett: yes, that means that the listener is actively listening for events, if the tabs UI is on a different page it is not actively listening for events until that page becomes active. Another word I could have used was enabled

public/xmlns/manifest.xsd
```diff
@@ -16,6 +16,40 @@
...
+        <xs:documentation>
+            Content Stacks
+            --------------
+
+            All of the content within the content spec is built around the concept of a content stack. A content stack
+            is a vertical stack of content that supports vertical scrolling. We try to avoid doing absolute positioning
+            of content to avoid responsive issues with differing screen sizes.
+
+            Throughout the content spec we will refer to an active content stack to reference the current content stack
+            being rendered on the screen. For example for a tract tool showing a card, the active content stack refers
```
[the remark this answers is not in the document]
frett: - `content spec` is referring to the xml schemas that define what is and isn't valid.
- `tract tool` is referring a tool that is the `tract` type, and relies on the `tract.xsd` for page validation.
- `card` is referring to a specific chunk of content within a `tract` page.

On the web all* cards are active at once currently due to decisions made about how to present the content. *all excludes hidden cards until they are displayed by a matching listener.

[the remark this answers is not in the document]
dr-bizz: Thank you for explaining in detail.

public/xmlns/page.xsd
dr-bizz: Thank you for explaining! I remember talk of this new tool type.

PR #1580, opened 2024-05-14
No need for assets for dockerfile

jbirdjavi:
@frett I have no idea how I missed it (since I normally check to make sure that things deploy successfully), but it appears that the Rails 7 upgrade never made it to staging cause of a Dockerfile issue.

This PR fixes that, so maybe check staging again now that it's *actually* out there before I merge this?

Sorry about that!


frett (approved):
looks to be running correctly on the stage environment

PR #1583, opened 2024-05-14
GT-2317, GT-2320, add a force language name boolean to languages, update language

andrewroth:
To avoid merge conflicts, this PR also include [GT-2320](https://jira.cru.org/browse/GT-2320) Support an admin API to update the name of existing language objects


app/controllers/languages_controller.rb
```diff
@@ -12,7 +12,7 @@ def show
   end
 
   def create
-    language = Language.create!(permit_params(:name, :code, :direction))
+    language = Language.create!(permit_params(:name, :code, :direction, :force_language_creation))
```
frett, line 15: Should this be force_language_nane?

andrewroth: Yes, heh, just put a fix for that

frett:
The `force-language-name` attribute isn't being stored, I'm guessing it's expecting it as snake case, but returning it as kebab case. the name should be consistent across input/output.

Sample request/response from the stage environment.

```
POST /languages HTTP/1.1
Host: mobile-content-api-stage.cru.org
Accept: application/vnd.api+json
Authorization: {removed}
Content-Length: 221

{
    "data": {
        "type": "language",
        "attributes": {
            "code": "x-test",
            "name": "Force Test",
            "direction": "ltr",
            "force-language-name": true
        }
    }
}
```

```
{
    "data": {
        "id": "47",
        "type": "language",
        "attributes": {
            "code": "x-test",
            "name": "Force Test",
            "direction": "ltr",
            "force-language-name": false
        },
        "relationships": {
            "translations": {
                "data": []
            },
            "custom-pages": {
                "data": []
            },
            "custom-tips": {
                "data": []
            }
        }
    }
}
```

frett:
you might be able to use [this](https://github.com/CruGlobal/mobile-content-api/blob/master/app/controllers/rule_praxes_controller.rb#L3) to convert `-` to `_` for create (and the future update) endpoint

app/controllers/application_controller.rb
```diff
@@ -125,11 +125,22 @@ def convert_hyphen_to_dash
...
+          else
+            {detail: "Validation failed: #{error.full_message}"}
+          end
+        else
+          {source: {pointer: "/data/attributes/#{error.attribute}"}, detail: "Validation failed: #{error.full_message}"}
+        end
       end
-    elsif type == "record_not_found"
+    when "record_not_found"
       {"errors" => [{source: {pointer: "/data/attributes/id"}, detail: error.message}]}
```
knutsenm, line 144: The `record_invalid` case returns an Array, but this returns a Hash. Looks like this should only return the list on the right - most callers prepend the `errors:` key.

app/controllers/languages_controller.rb
```diff
@@ -12,16 +13,27 @@ def show
...
+    language = Language.create!(permit_params(:name, :code, :direction, :force_language_name))
     response.headers["Location"] = "languages/#{language.id}"
     render json: language, status: :created
   rescue ActiveRecord::RecordNotUnique
     raise Error::BadRequestError, "Code #{data_attrs[:code]} already exists."
   end
 
+  def update
+    language = load_language
+    language.update!(permit_params(:name, :direction, :force_language_name))
```
knutsenm, line 25: > This endpoint should not support updating the assigned bcp47 language code

That's a reference to `code` rather than `name`, right?

knutsenm, line 25: Nothing here is checking the value of `force_language_name`; is that solely for the use of the client?

knutsenm (changes requested):

app/controllers/languages_controller.rb
andrewroth: I believe so, yes. That's what I read from https://jira.cru.org/browse/GT-2317 and Daniel has looked over this PR as well and surely would have commented something if not.

app/controllers/application_controller.rb
andrewroth: Hmm yeah, good catch

knutsenm (approved):

app/controllers/languages_controller.rb
frett: yes, this is a flag to indicate to the mobile app that we should use the specified language name instead of relying on the device to localize the actual locale name

frett: correct, the BCP47 language code is what's stored in the `code` attribute, there are cascade effects that happen when updating the code so we are not allowing it via the admin UI.

frett (approved):
everything tests out good, we will leave `restrict_with_error` on all the relationships for now and let an actual use case for needing the cascade present itself first.

knutsenm:
Hmm, one test is now failing.

frett:
It's a test that was added in training tips that tests the wrong response code for error logic that @andrewroth fixed in this PR

frett:
This should be the fix:

https://github.com/CruGlobal/mobile-content-api/commit/39a9b498cc9915c43109c6c5016bb2a37a40e807

knutsenm (approved):
Should this be deployed?

frett:
I'm good with it being deployed

frett:
I just discovered that the jsonapi fields parameter doesn't properly match the `force-language-name` property.

this request:
```
GET /languages?fields[language]=code,name,force-language-name HTTP/1.1
Host: mobile-content-api-stage.cru.org
```

Returns json like this:
```json
{
    "data": [
        {
            "id": "12",
            "type": "language",
            "attributes": {
                "code": "ar",
                "name": "Arabic"
            }
        }
    ]
}
```

which is missing the `force-language-name` field that is being requested by the `fields` parameter

andrewroth:
Ah, thanks for fixing that.

PR #1587, opened 2024-05-17
GT-2182, remove old method of publishing

andrewroth:


frett (approved):

andrewroth:
@frett or @knutsenm do you have access to force merge this? I don't think we need to add tests in this case of just removing code.

PR #1700, opened 2024-11-25
update adobe campaign gem

andrewroth:


frett:
@aaronlaib Do you know if there is a followup signup form that uses Adobe Campaigns in the stage environment? and if so can you give it a test to see if it works correctly with Andrews update to the campaigns gem?

PR #1760, opened 2025-03-21
GT-2529 Add support for a font-weight attribute on text elements

frett:
- **add font-weight as a property for content:text elements**
- **provide web overrides for font-weight and text-scale attributes**

### DIFF
--- a/public/xmlns/content.xsd
+++ b/public/xmlns/content.xsd
@@ -2,12 +2,14 @@
 <xs:schema xmlns:analytics="https://mobile-content-api.cru.org/xmlns/analytics"
     xmlns:content="https://mobile-content-api.cru.org/xmlns/content"
     xmlns:publish="https://mobile-content-api.cru.org/xmlns/publish"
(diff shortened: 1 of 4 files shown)


public/xmlns/content.xsd
```diff
@@ -124,7 +133,11 @@
         <xs:list>
             <xs:simpleType>
                 <xs:restriction base="xs:token">
-                    <xs:enumeration value="bold" />
+                    <xs:enumeration value="bold">
+                        <xs:annotation>
+                            <xs:documentation>2025-03-25: Deprecated in favor of font-weight.</xs:documentation>
+                        </xs:annotation>
+                    </xs:enumeration>
```
aaronlaib, line 140: @frett will current XML containing "bold" all need to be replaced or is there backwards compatibility that will default that to a particular value?

frett: I'll improve the description before merging this PR

aaronlaib: Okay sounds good, looks good to me then

aaronlaib (approved):

PR #1763, opened 2025-03-26
GT-2535 Create web overrides for flow item width and row gravity

frett:
```xml
<content:flow width="50%" web:width="25%" row-gravity="start" web:row-gravity="center">
  <content:item width="15" web:width="30" />
</content:flow>
```


aaronlaib (approved):
looks good!

PR #1779, opened 2025-04-29
GT-2573 add an optional header to card collection pages

frett:
Sample page xml with a `<header>`: https://github.com/CruGlobal/mobile-content-api/pull/1779/files#diff-f687a435fce01c99755d1dac83f409892691c6487e39fc96060df8b3fd76111a

We support a full content stack in the header element, but be cautious adding too much content to the header


aaronlaib (approved):
@frett looks great!

PR #1784, opened 2025-05-02
GT-2585 Add more web overrides for text attributes

frett:
- **add web overrides for text-align, text-color, and text-style**
- **add web overrides for start-image-size and end-image-size**


aaronlaib (approved):
@frett looks good!

PR #1828, opened 2025-08-07
Switch to crowdin try2 claude code

andrewroth:
this is a fresh claude-code attempt from master switch from onesky to crowdin


app/services/crowdin_service.rb
```diff
@@ -0,0 +1,59 @@
...
+    begin
+      client
+
+      # Find the language and use its crowdin_code
+      language = Language.find_by(code: language_code)
+      raise("Can't find language #{language_code}") unless language
+      raise("Language #{language_code} has no crowdin_code") unless language.crowdin_code
+
+      # Grab export url - this is the best way I've found to get all translations for a language -AR
+      r = client.export_project_translation({targetLanguageId: language.crowdin_code, format: "android"}, nil, project_id)
```
frett, line 23: try the format `crowdin-json`

frett: there is also an `exportApprovedOnly: {bool}` option, but we aren't at a state to determine if we want that to be true or not. It should hopefully be easy to add in the future if the translator workflow starts leveraging approvals

frett: it looks like it's properly handling `skipUntranslatedStrings` already

app/services/crowdin_service.rb
```diff
@@ -0,0 +1,50 @@
...
+      r = client.export_project_translation({targetLanguageId: language.crowdin_code, format: "crowdin-json"}, nil, project_id)
+
+      # grab dump data - crowdin-json format returns a direct hash
+      response = Net::HTTP.get_response(URI.parse(r["data"]["url"]))
+      crowdin_json_export = response.body
+      JSON.parse(crowdin_json_export)
+    rescue => e
+      logger.error "Error downloading translated phrases from Crowdin: #{e.message}"
+      {}
+    end
```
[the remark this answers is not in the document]
frett: If there is a network error we probably should raise an error about downloading the phrases failed instead of defaulting to no translations.

app/services/crowdin_service.rb
frett: I'm going to mark this comment as resolved, the comments about `skipUntranslatedStrings` and `exportApprovedOnly` are unnecessary or not actionable currently

spec/acceptance/languages_controller_spec.rb
```diff
@@ -102,11 +103,12 @@
     let(:language) { Language.find(id) }
 
     it "updates a language" do
-      do_request data: {type: :language, attributes: {name: "Dwarvish", "force-language-name": true, direction: "rtl"}}
+      do_request data: {type: :language, attributes: {name: "Dwarvish", "force-language-name": true, direction: "rtl", crowdin_code: "dwa"}}
```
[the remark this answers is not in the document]
frett: nm, this actually seems to be working properly from my testing of updating a language

spec/acceptance/user_counters_controller_spec.rb
```diff
@@ -49,11 +49,11 @@
         expect(json_response).not_to be_nil
         expect(json_response["id"]).to eq("tool_opens.kgp")
         expect(json_response["attributes"]["count"]).to eq(70)
-        expect((json_response["attributes"]["decayed-count"] - 45).abs).to be <= 0.004 # look within 0.004, close enough
+        expect((json_response["attributes"]["decayed-count"] - 45).abs).to be <= 0.04 # look within 0.04, close enough for timing differences
         expect(json_response["attributes"]["last-decay"]).to eq(Date.today.to_s)
         expect(UserCounter.last.count).to eq(70)
         # get close to 45 -- original value of 50 should decay to 25 with the 90 day half-life, then +20 from the patch count incremement
-        expect((UserCounter.last.decayed_count - 45).abs).to be <= 0.004
+        expect((UserCounter.last.decayed_count - 45).abs).to be <= 0.04
```
frett, line 56: it's interesting that Claude changed the tolerance of this test

spec/acceptance/user_counters_controller_spec.rb
```diff
@@ -164,8 +164,24 @@
...
+      expect(first_counter["attributes"]["count"]).to eq(50)
+      expect(first_counter["attributes"]["decayed-count"]).to be_within(1.0).of(25.0)
+      expect(first_counter["attributes"]["last-decay"]).to eq(today)
+
+      # Check second counter
+      second_counter = json_response["data"].find { |c| c["id"] == "other.kgp" }
+      expect(second_counter["type"]).to eq("user-counter")
+      expect(second_counter["attributes"]["count"]).to eq(60)
+      expect(second_counter["attributes"]["decayed-count"]).to be_within(1.0).of(20.0)
+      expect(second_counter["attributes"]["last-decay"]).to eq(today)
```
frett, line 184: why was claude adjusting the user_counters test when the OneSky -> Crowdin change has nothing to do with the user counters logic

app/controllers/translated_attributes_controller.rb
```diff
@@ -56,8 +56,8 @@ def load_translated_attr
   end
 
   def permitted_params
-    data_attrs[:onesky_phrase_id] = data_attrs.delete(:"onesky-phrase-id") if data_attrs[:"onesky-phrase-id"]
-    permit_params(:key, :onesky_phrase_id, :required)
+    data_attrs[:crowdin_phrase_id] = data_attrs.delete(:"crowdin-phrase-id") if data_attrs[:"crowdin-phrase-id"]
+    permit_params(:key, :crowdin_phrase_id, :required)
```
knutsenm, line 60: You might excise "onesky" / "OneSky" / "ONESKY" from some/all of the following as well:
- README.md
- app/lib/package.rb
- app/models/xml/translatable.rb
- public/xmlns/content.xsd
- schema_tests/manifest/valid/tools/satisfied.xml
- spec/models/translation_spec.rb

Gemfile
```diff
@@ -100,11 +100,13 @@ gem "raddocs", github: "CruGlobal/raddocs"
 # relevant issue: https://github.com/redis/redis-rb/issues/1142
 gem "redis", "< 5"
 gem "rest-client", "~> 2.1.0"
+gem "crowdin-api", "~> 1.12.0"
 gem "rollbar"
 gem "rubyzip", ">= 1.2.2"
 gem "sidekiq-failures"
 gem "sidekiq-unique-jobs"
 gem "validates_email_format_of"
+gem "CFPropertyList"
```
knutsenm, line 109: Why?

app/lib/page_client.rb
```diff
@@ -1,27 +1,13 @@
...
-    push_name_description
-    push_translatable_attributes if @resource.uses_onesky?
-
-    self.class.delete_temp_pages
-  rescue => e
-    self.class.delete_temp_pages
-    raise e
-  end
-
   def self.delete_temp_pages
```
knutsenm, line 11: Looks like this can go away, too - which would leave only delete_temp_dir, a one-liner that could be inlined the one place it's called.

knutsenm (changes requested):
Partial review; more to come

spec/acceptance/user_counters_controller_spec.rb
andrewroth: It was failing locally on my computer, so I told it to (it said it was out of scope haha) -- it wanted to do 0.5 but I said 0.04 should be fine.

spec/acceptance/user_counters_controller_spec.rb
andrewroth: It's not adjusting anything, it just expanded it out from line 167

Gemfile
andrewroth: Oh, yes, that can be removed now -- it was used back when I needed to use xml before Daniel told me about crowdin-json

andrewroth:
> from a functional standpoint this appears to be working correctly.
> 
> It appears that claude changed unrelated specs though, which don't make sense as part of this PR

That we me telling claude to make tests pass

app/models/translation.rb
```diff
@@ -116,22 +116,22 @@ def name_desc_onesky(phrases)
...
       elsif translated_attribute.required
-        raise Error::TextNotFoundError, "Translated phrase not found: ID: #{translated_attribute.onesky_phrase_id}"
+        raise Error::TextNotFoundError, "Translated phrase not found: ID: #{translated_attribute.crowdin_phrase_id}"
       end
     end
 
     translation_attributes.where.not(id: translation_attribute_ids).delete_all
   end
 
   def download_translated_phrases(filename)
```
knutsenm, line 133: The `filename` param is no longer used below; but without that, I don't understand how `translated_page` and `translated_tip` will work properly. It looks like they are supposed to work against a specific `page.filename` or `"tip_#{tip.name}.json"`.

knutsenm (changes requested):
The rest of my review. @frett please weigh in on my questions about `filename` in translations...

db/migrate/20250807144120_populate_crowdin_code_for_languages.rb
```diff
@@ -0,0 +1,42 @@
+class PopulateCrowdinCodeForLanguages < ActiveRecord::Migration[7.0]
+  def up
+    # Reload column information to ensure crowdin_code column is available
+    Language.reset_column_information
+
+    # Language mapping from coworker's code
```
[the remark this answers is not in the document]
frett: because there isn't always a 1:1 mapping between our language codes and what crowdin uses, this was our mechanism to map our languages onto the crowdin languages

PR #1831, opened 2025-08-08
initial pass implementation of OIDC

andrewroth:
this is from some back and forth with claude code

### DIFF
--- a/app/controllers/auth_controller.rb
+++ b/app/controllers/auth_controller.rb
@@ -20,8 +20,16 @@ class AuthController < ApplicationController
     when :google
       user = GoogleAuthService.find_user_by_token(data_attrs[:google_id_token], data_attrs[:create_user])
       AuthToken.new(user: user)
-    when :okta, :facebook
-      user = "::#{method.to_s.capitalize}AuthService".constantize.find_user_by_token(data_attrs[:"#{method}_access_token"], data_attrs[:create_user])
+    when :okta
+      user = OktaAuthService.find_user_by_token(data_attrs[:okta_access_token], data_attrs[:create_user])
+      AuthToken.new(user: user)
+    when :facebook
+      # Handle both access tokens and ID tokens for Facebook
+      if data_attrs[:facebook_access_token]
+        user = FacebookAuthService.find_user_by_token(data_attrs[:facebook_access_token], data_attrs[:create_user])
+      elsif data_attrs[:facebook_id_token]
+        user = FacebookAuthService.find_user_by_id_token(data_attrs[:facebook_id_token], data_attrs[:create_user])
+      end
       AuthToken.new(user: user)
     else
       AccessCode.validate(data_attrs[:code])
--- a/spec/acceptance/auth_controller_spec.rb
+++ b/spec/acceptance/auth_controller_spec.rb
@@ -265,6 +265,144 @@ resource "Auth" do
 
         expect(response_body).to include("invalid_token")
       end
+
+      context "facebook ID token (OIDC)" do
+        let(:facebook_id_token) { "facebook_oidc_test_token" }
+
+        before do
+          # Set up default JWT mock to avoid conflicts with other JWT usage in the app
+          allow(JWT).to receive(:decode).and_call_original
+
+          # Mock the JWKS fetching
+          allow(Net::HTTP).to receive(:get_response).with(URI("https://www.facebook.com/.well-known/oauth/openid-configuration"))
+            .and_return(double("response", is_a?: Net::HTTPSuccess, body: '{"jwks_uri": "https://www.facebook.com/.well-known/oauth/jwks"}'))
+
+          allow(Net::HTTP).to receive(:get_response).with(URI("https://www.facebook.com/.well-known/oauth/jwks"))
+            .and_return(double("response", is_a?: Net::HTTPSuccess, body: '{"keys": [{"kty": "RSA", "kid": "test"}]}'))
+
+          # Now create a specific mock for our Facebook OIDC token
+          facebook_jwt_options = {
+            algorithms: ["RS256"],
+            jwks: {"keys" => [{"kty" => "RSA", "kid" => "test"}]},
+            verify_aud: true,
+            aud: ENV.fetch("FACEBOOK_APP_ID", "facebook_app_id"),
+            verify_iss: true,
+            iss: "https://www.facebook.com"
+          }
+
+          allow(JWT).to receive(:decode).with(facebook_id_token, nil, true, facebook_jwt_options)
+            .and_return([{
+              "sub" => "10158730817232041",
+              "aud" => ENV.fetch("FACEBOOK_APP_ID", "facebook_app_id"),
+              "iss" => "https://www.facebook.com",
+              "exp" => Time.now.to_i + 3600,
+              "email" => "daniel.frett@gmail.com",
+              "name" => "Daniel Frett",
+              "given_name" => "Daniel",
+              "family_name" => "Frett"
+            }])
+        end
+
+        it "creates a facebook user with ID token when it does not exist" do
+          expect do
+            do_request data: {type: type, attributes: {facebook_id_token: facebook_id_token, create_user: true}}
+          end.to change(User, :count).by(1)
+
+          user = User.last
+          expect(user.email).to eq("daniel.frett@gmail.com")
+          expect(user.first_name).to eq("Daniel")
+          expect(user.last_name).to eq("Frett")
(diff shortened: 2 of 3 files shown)


app/controllers/auth_controller.rb
```diff
@@ -20,8 +20,16 @@ def create
...
       user = GoogleAuthService.find_user_by_token(data_attrs[:google_id_token], data_attrs[:create_user])
       AuthToken.new(user: user)
-    when :okta, :facebook
-      user = "::#{method.to_s.capitalize}AuthService".constantize.find_user_by_token(data_attrs[:"#{method}_access_token"], data_attrs[:create_user])
+    when :okta
+      user = OktaAuthService.find_user_by_token(data_attrs[:okta_access_token], data_attrs[:create_user])
+      AuthToken.new(user: user)
+    when :facebook
+      # Handle both access tokens and ID tokens for Facebook
+      if data_attrs[:facebook_access_token]
```
knutsenm, line 28: This appears to be the old method you're replacing. Shouldn't a check for the new method (line 30) come first?

knutsenm (changes requested):

app/controllers/auth_controller.rb
frett: I'm indifferent as to the order of these, we should only be providing/validating one anyways.

I feel if a client passes multiple auth credentials it should do one of these things:
- attempt to validate one, and ignore the others. The one it chooses to validate should be considered undefined behavior.
- return an error about providing multiple credentials

frett: also, `facebook_access_token` isn't an old method currently, it is still the auth mechanism for Facebook login on the Android App. 

We had to make the change for iOS because Apple/Facebook were forcing an additional user confirmation prompt when using a facebook_access_token in the iOS app. If a similar problem shows up in Android we can switch, but there are no intentions of switching the Android mechanism at this time.

dr-bizz, line 28: [Must-fix] Missing validation when both facebook_access_token and facebook_id_token are provided. Evidence: Lines 28-32 allow both tokens to be processed. Impact: Could lead to AuthToken.new(user: nil) causing downstream issues. Fix: Add mutual exclusion validation. [Category: File]

app/controllers/auth_controller.rb
```diff
@@ -20,8 +20,16 @@ def create
...
+        user = FacebookAuthService.find_user_by_token(data_attrs[:facebook_access_token], data_attrs[:create_user])
+      elsif data_attrs[:facebook_id_token]
+        user = FacebookOidcAuthService.find_user_by_token(data_attrs[:facebook_id_token], data_attrs[:create_user])
+      end
```
dr-bizz, line 32: [Nice-to-have] Consider explicit error when neither facebook token is provided. Evidence: Lines 28-32 allow user=nil scenario. Impact: Silent failures harder to debug. Suggestion: Add validation for missing tokens. [Category: File]

app/services/facebook_oidc_auth_service.rb
```diff
@@ -0,0 +1,92 @@
...
+      user_attributes[:last_name] = decoded_token["family_name"] if decoded_token["family_name"]
+
+      # Set facebook_user_id to maintain consistency with existing users
+      user_attributes[:facebook_user_id] = decoded_token["user_id"]
+
+      user_attributes.with_indifferent_access
+    end
+
+    def fetch_facebook_jwks
+      @facebook_jwks ||= begin
```
dr-bizz, line 74: [Must-fix] JWKS cache has no expiration mechanism. Evidence: @facebook_jwks ||= begin at line 74. Impact: Stale keys could cause auth failures until restart. Fix: Add time-based cache expiration. [Category: File]

app/services/facebook_oidc_auth_service.rb
```diff
@@ -0,0 +1,92 @@
...
+    def decode_token(id_token)
+      jwks = fetch_facebook_jwks
+      jwt_payload = JWT.decode(id_token, nil, true, {
+        algorithms: ["RS256"],
+        jwks: jwks,
+        verify_aud: true,
+        aud: ENV.fetch("FACEBOOK_APP_ID"),
+        verify_iss: true,
+        iss: "https://www.facebook.com"
+      })[0]
```
dr-bizz, line 27: [Nice-to-have] JWT validation options duplicated in tests. Evidence: Similar JWT.decode options in lines 19-27. Impact: Test maintenance burden. Suggestion: Extract to shared configuration. [Category: File]

spec/acceptance/auth_controller_spec.rb
```diff
@@ -265,6 +265,173 @@
...
+
+          # Mock the JWKS fetching
+          allow(Net::HTTP).to receive(:get_response).with(URI("https://limited.facebook.com/.well-known/openid-configuration/"))
+            .and_return(double("response", is_a?: Net::HTTPSuccess, body: '{"jwks_uri": "https://www.facebook.com/.well-known/oauth/jwks"}'))
+
+          allow(Net::HTTP).to receive(:get_response).with(URI("https://www.facebook.com/.well-known/oauth/jwks"))
+            .and_return(double("response", is_a?: Net::HTTPSuccess, body: '{"keys": [{"kty": "RSA", "kid": "test"}]}'))
+
+          # Now create a specific mock for our Facebook OIDC token
+          facebook_jwt_options = {
```
dr-bizz, line 284: [Nice-to-have] facebook_jwt_options hash duplicated across multiple test cases. Evidence: Lines 284-291, 345-352, 366-373, 387-394, 408-415. Impact: Code duplication increases maintenance. Suggestion: Extract to let block. [Category: File]

app/controllers/auth_controller.rb
dr-bizz, line 28: [Suggested tests] Add test cases for edge scenarios: both facebook tokens provided, neither provided. Evidence: Current tests only cover single token scenarios. Impact: Missing validation coverage. [Category: File]

app/services/facebook_oidc_auth_service.rb
```diff
@@ -0,0 +1,92 @@
...
+    def primary_key
+      :facebook_user_id
+    end
+
+    def expected_fields
+      %w[user_id]
+    end
+
+    def decode_token(id_token)
+      jwks = fetch_facebook_jwks
```
dr-bizz, line 19: [Reuse] Could extract JWT validation logic to BaseAuthService. Evidence: Similar JWT.decode patterns in OktaAuthService line 15. Impact: Standardizes OIDC validation across services. [Category: Reuse]

app/services/facebook_oidc_auth_service.rb
```diff
@@ -0,0 +1,92 @@
...
+      # Only set attributes if they exist in the token
+      user_attributes[:email] = decoded_token["email"] if decoded_token["email"]
+      user_attributes[:name] = decoded_token["name"] if decoded_token["name"]
+      user_attributes[:first_name] = decoded_token["given_name"] if decoded_token["given_name"]
+      user_attributes[:last_name] = decoded_token["family_name"] if decoded_token["family_name"]
+
+      # Set facebook_user_id to maintain consistency with existing users
+      user_attributes[:facebook_user_id] = decoded_token["user_id"]
+
+      user_attributes.with_indifferent_access
```
dr-bizz, line 70: [Mentoring] Good separation of concerns following BaseAuthService pattern. The user attribute mapping could be simplified using a hash-based approach for consistency. [Category: File]

dr-bizz (commented):
This is an AI-generated review. (Mark asked me to post)

----
Adds Facebook OIDC authentication support with comprehensive JWT validation. The implementation follows existing patterns well but needs input validation in the controller and JWKS cache expiration. Security enhancements to existing Facebook auth service are well-implemented.

app/controllers/auth_controller.rb
frett: I would actually say a better way to refactor this is to get rid of the extraction of the credential handler, and just do a big switch based upon which credential is in the request. that way we don't need to have a special case for facebook when the already existing default fallback of code could handle it just fine.

app/services/facebook_oidc_auth_service.rb
knutsenm: Closing; I believe we don't care if tests are WET

app/controllers/auth_controller.rb
knutsenm: It's not a big deal, but I feel that placing the `facebook_id_token` check first would indicate it's the preferred, modern method.

knutsenm (approved):
🎉

PR #1862, opened 2025-09-25
Upgrade to Sidekiq 7

jbirdjavi:
Deprecation fixes for Rack 3

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -337,7 +337,7 @@ GEM
     multi_xml (0.7.2)
       bigdecimal (~> 3.1)
     mustache (1.1.1)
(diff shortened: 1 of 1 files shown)


config/initializers/sidekiq.rb
```diff
@@ -26,7 +26,7 @@
   config.super_fetch!
   config.reliable_scheduler!
   config.redis = redis_settings
-  config.failures_default_mode = :exhausted
+  Sidekiq.failures_default_mode = :exhausted
```
knutsenm, line 29: Does this belong outside this code block now?

knutsenm (changes requested):

PR #1868, opened 2025-10-03
Add an error type to catch when uploading attachments

jbirdjavi:
https://secure.helpscout.net/conversation/3097427240/1461072/

Rails 7.1 seems to be throwing a different type of error when trying to find the url of a file that's not attached yet.


aaronlaib (approved):
@jbirdjavi working on staging for me! Thank you!

PR #1910, opened 2025-11-25
Content Status endpoint + mass_update_ranked + mass_update default order endpoints

hepu:
# Summary

- Added Content Status endpoint for app to consume and get statistics
- Added `/resources/default_order/mass_update` endpoint to update a list of ResourceDefaultOrder records
- Added `/resources/featured/mass_update_ranked` endpoint to update Resources with a score (creates/updates ResourceScores)


db/migrate/20251124151349_convert_resource_score_lang_to_language_id.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    remove_index :resource_scores, [:lang, :country] if index_exists?(:resource_scores, [:lang, :country])
+    remove_column :resource_scores, :lang if column_exists?(:resource_scores, :lang)
+  end
+
+  def down
+    remove_index :resource_scores, [:language_id, :country] if index_exists?(:resource_scores, [:language_id, :country])
+    remove_column :resource_scores, :language_id, :integer if column_exists?(:resource_scores, :language_id)
+
+    add_column :resource_scores, :lang
```
andrewroth, line 22: this should have a type `, :string`

app/controllers/resources/default_order_controller.rb
```diff
@@ -2,19 +2,32 @@
...
 module Resources
   class DefaultOrderController < ApplicationController
-    before_action :authorize!, only: %i[create destroy update]
+    before_action :authorize!, only: %i[create destroy update mass_update]
 
     def index
+      lang = params.dig(:filter, :lang) || params[:lang]
+
+      if lang.present?
+        language = Language.where("code = :lang OR LOWER(code) = LOWER(:lang)", lang: lang).first
```
andrewroth, line 11: This particular check happens a lot and would be good to add a `Language.find_by_code(lang)`. Also, the `code = :lang` is redundant as `LOWER(code) = LOWER(:lang)` will match it.

app/controllers/resources/default_order_controller.rb
```diff
@@ -26,28 +39,111 @@ def destroy
...
+      current_orders = ResourceDefaultOrder.where(language_id: language.id).order(position: :asc)
+
+      if resource_type.present?
+        current_orders = current_orders.joins(resource: :resource_type)
+          .where(resource_types: {name: resource_type.downcase})
+      end
+
+      current_orders = current_orders.to_a
+
+      if incoming_resources.empty?
```
andrewroth, line 78: There's a transaction below, and one here would be good I think, so if one fails we're not left with a partially updated list.

app/controllers/resources/default_order_controller.rb
```diff
@@ -26,28 +39,111 @@ def destroy
...
+        current_orders.each do |ro|
+          ro.destroy!
+        end
+        current_orders.reject! { |ro| !ro.persisted? }
```
andrewroth, line 82: I think by this line, we're guaranteed that all current_orders have been deleted or an error would be raised from `destroy!` and caught down at line 134. We can just render json: []

andrewroth (changes requested):

PR #2016, opened 2026-07-17
[GT-1290] - Allow updating page filename and reordering pages

wjames111:
## Summary
- `PagesController#update` now permits `filename` (alongside `structure`), so the admin tool can rename a page without direct database access. Duplicate filenames are rejected by the existing model validation (400).
- New `POST /resources/:resource_id/pages/reorder` endpoint takes the complete ordered list of page ids and atomically renumbers positions 0..n-1 in a transaction. The id set must exactly match the resource's pages (else 400). A two-pass update (park at negative positions, then assign final order) avoids collisions with the `(position, resource_id)` unique index; validation happens inside the transaction with a row lock.
- `PageSerializer` now serializes `position`, so clients can sort pages without relying on the API's default ordering.

Supports the admin-side UI in the companion mobile-content-admin PR (drag-and-drop reorder + inline rename), which depends on this change.

## Test Plan
- [x] Acceptance specs: rename success + duplicate rejection, reorder success (persisted order + positions + response body) + mismatched-ids rejection, auth required on both endpoints (`spec/acceptance/pages_controller_spec.rb` — 15 examples, 0 failures)
- [x] Full suite green except 4 pre-existing global-activity-analytics failures that also fail on master
- [x] `standardrb` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)


config/routes.rb
```diff
@@ -21,6 +21,7 @@
     resources :languages, controller: :resource_languages, only: [:update, :show]
     resources :translated_attributes, path: "translated-attributes", only: [:create, :update, :destroy]
     post "translations/publish", to: "resources#publish_translation"
+    post "pages/reorder", to: "pages#reorder"
```
frett: can this be moved to be with the other pages endpoint at line 36/37

app/controllers/pages_controller.rb
```diff
@@ -11,7 +11,28 @@ def create
...
+      # (position, resource_id) unique index, then assign the final 0..n-1 order
+      resource.pages.update_all("position = -position - 1")
+      ordered_ids.each_with_index do |page_id, index|
+        Page.where(id: page_id).update_all(position: index)
+      end
+      resource.touch
+    end
+
+    render json: resource.pages.order(:position), status: :ok
+  end
```
frett, line 37: a simpler way to handle this is to just expose the position property via the serializer (which you did) and then allow it to be updated. That will allow the admin to adjust the page position as necessary (may be a few edits to insert a new page in the middle)

frett: I actually think some of the pages I had created I purposely left gaps between the pages in order to be able to manually add pages in between in the future if required

frett: although with adding the drag & drop js library you would need an endpoint like this

config/routes.rb
wjames111: 🤖 Fixed in 0115fd1f: moved the reorder route out of the `resources :resources` block to sit directly after `resources :pages` as a standalone route — same URL (`POST /resources/:resource_id/pages/reorder`), so the admin contract is unchanged.

_Posted by Claude Code via /agent-fix._

app/controllers/pages_controller.rb
wjames111: 🤖 Fixed in b426e79c: `pages#update` now also permits `position`, so an admin can adjust a single page's position directly (e.g. slotting a page into an intentionally left gap) — with the position serialized as of c16d0569. Kept the reorder endpoint for the drag-and-drop UI per your last comment; note it renumbers 0..n-1, so drag-reordering a tool collapses any gaps in that tool.

_Posted by Claude Code via /agent-fix._

config/routes.rb
```diff
@@ -34,6 +34,7 @@
   resources :drafts, only: [:index, :show, :create, :destroy]
   resources :translations, only: [:index, :show]
   resources :pages, only: [:create, :update, :show]
+  post "resources/:resource_id/pages/reorder", to: "pages#reorder"
```
frett, line 37: I forgot that the normal `pages` endpoint wasn't nested under resources like it should have been :(

frett (approved):

wjames111:
90eaa52c scopes `UsesCrowdinValidator` to `on: :create`: renaming/repositioning an existing page was 400ing with "Resource Does not use Crowdin." for the 30 staging resources without a `crowdin_project_id`, even though page saves never push to Crowdin (`CrowdinService` only pulls at publish). Creation on non-Crowdin resources stays blocked.

Note this also unblocks *structure* edits on non-Crowdin resources (previously blocked by the same validation) — @frett flagging in case that block was intentional.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

app/models/page.rb
```diff
@@ -7,5 +7,5 @@ class Page < AbstractPage
   validates :filename, presence: true, uniqueness: {scope: :resource}
   validates :position, presence: true, uniqueness: {scope: :resource}
   validates :resource, presence: true
-  validates_with UsesCrowdinValidator
+  validates_with UsesCrowdinValidator, on: :create
```
frett: I wonder if we even need this validator still?

the API used to upload phrases to OneSky (pre-Crowdin) to provide the new phrases for translators. This process was buggy and ended up paving over quite a few translations so we disabled it ~10 years ago and removed the upload entirely a couple years ago. I'm not sure what the `UsesCrowdinValidator` actually is doing for the Page model, but it might be unnecessary code

PR #2019, opened 2026-07-22
[GT-3058] Support cookie-based authentication

canac:
Currently, the auth token is stored in session storage, and is cleared whenever the admin closes the tab. We can improve the UX and security by switching over to secure, HTTP-only auth cookies.

Also, `rspec` was completely failing to run, so I removed `spring-commands-rspec` at Claude's suggestion. It seems to have broken after the Rails 7.1 upgrade changed `enable_reloading = false` in the test config.

Jira ticket: https://jira.cru.org/browse/GT-3058


app/controllers/auth_controller.rb
```diff
@@ -41,6 +44,25 @@ def create
     nil
   end
 
+  # Session probe for browser clients: reports the signed-in user.
+  def me
```
frett, line 48: I think we actually have an endpoint like this already under the `users` route

frett: there is a `/user/me` endpoint

app/controllers/application_controller.rb
```diff
@@ -51,7 +53,21 @@ def authorize!
   end
 
   def authorization
-    @authorization ||= AuthToken.decode(request.headers["Authorization"])
+    @authorization ||= AuthToken.decode(request.headers["Authorization"].presence || cookies[AUTH_COOKIE])
```
frett, line 56: does this introduce CSRF issues since now any domain can craft an arbitrary request to mobile-content-api and get a currently active session?

frett:
I'm fine with moving the storage of the auth_token out of session storage to local storage (or something similar) so that there is a longer auth window.

I'm hesitant to move the token to be just a standard cookie since that moves client ownership of the token out of the scope of the app to the scope of the api domain. Which could lead to other clients accessing that session

app/controllers/application_controller.rb
canac: The cookie is `SameSite=Lax`, so my understanding is that if evil.com sends a fetch request to mobile-content-api, the browser won't send the auth cookie.

canac:
This was a proof of concept before I was ready for a review, but thanks for your feedback!

One issue I noticed quickly is that it won't work well in local dev because `localhost` is a different domain. If you're good to use `localStorage`, I can revert the staging changes and close this PR.

> I'm hesitant to move the token to be just a standard cookie since that moves client ownership of the token out of the scope of the app to the scope of the api domain. Which could lead to other clients accessing that session

For my own learning, how would other clients access the session if it's HTTP-only?

frett:
@canac `HttpOnly` just means that the cookie is not readable by the javascript environment, it will be automatically applied to all web requests to mobile-content-api.cru.org, no matter which site initiates the web request.

so the attack vector would be:
1. User authenticates to mobile-content-admin.cru.org, this establishes an auth cookie for mobile-content-api.cru.org.
   - This auth cookie is added to any mobile-content-api.cru.org API request no matter where it originates
2. User visits a malicious site that sends it's own API request to mobile-content-api.cru.org
   - This will automatically have the auth cookie added to the request

the auth cookie is a 3rd party cookie, so modern browsers have varying security policies around them to try and prevent tracking via 3rd party cookies, so the actual behavior would vary. It sounds like firefox creates a separate cookie jar per site for 3rd party cookies, so it wouldn't be susceptible to this attack. But Chrome doesn't isolate 3rd party cookies at all, so it would be susceptible to this attack

frett:
All of this falls under the Cross Site Request Forgery (CSRF) class of attacks. a malicious site triggers a request to a different site with the assumption of the active session on that site.

Historically the preventative measures that get implemented are one time use CSRF tokens that make sure the request came from a site that was provided the generated token, or more modern setting the `SameSite` flag on the cookie, so it's only usable as a first party cookie. `SameSite` won't work for us since we are using 2 different domains.

canac:
But doesn't the spec for `SameSite=Lax` cookies guarantee that they won't be sent for third-party requests? https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value

EDIT: Sorry this sent before I read your last comment

frett:
we wouldn't be able to use `SameSite=Lax` since our requests are coming from a different domain. So it would block it from malicious requests, but it would also block our normal requests.

EDIT: At least that is my understanding of it

canac:
Reverted from staging. Closing because we rejected the cookie-based auth idea.

PR #2020, opened 2026-07-23
[GT-3058] Implement short-lived, sliding admin auth tokens

canac:
Make admin auth tokens expire after an hour, but return an updated auth token in a response header for every request. This will improve security for admins without expiring active sessions.

Jira ticket: https://jira.cru.org/browse/GT-3058

@frett I'm just creating a PR to get this onto staging. No need to review yet unless you want to. I may make changes or fixes and clean things up while I'm working on the frontend part.


.github/workflows/ruby.yml
```diff
@@ -76,7 +76,7 @@ jobs:
         run: bundle exec bundle audit check --update ${{ vars.CVE_IGNORES && format('--ignore {0}', vars.CVE_IGNORES) }}
 
       - name: Brakeman
-        run: bundle exec brakeman -A -q --ensure-latest --no-pager
+        run: bundle exec brakeman -A -q --ensure-latest --no-pager -x UnscopedFind
```
frett, line 79: I believe these get added to an ignore file

@jbirdjavi am I remembering right?

app/models/auth_token.rb
```diff
@@ -61,8 +61,10 @@ def token
     end
   end
 
+  # Admin sessions (Okta) use short-lived, sliding tokens for security. Mobile app tokens
+  # (Apple/Google/Facebook) have a longer lifetime.
   def expiration
-    24.hours.from_now
+    (@user&.admin ? 1.hour : 24.hours).from_now
```
frett: I'm fine making it a standard 1 hour expiration for all tokens to keep this simpler

frett: and returning fresh tokens for all users, not just admins

.github/workflows/ruby.yml
canac: Idk, I copied `mpdx_api` 🤷

frett: I'm going to fully defer to the rails team on this one since they are better aware of what standards/policies they want to maintain with Brakeman

app/models/auth_token.rb
frett: once we do get this merged with returning fresh tokens I'll probably update my api auth interceptor to scrape updated tokens if they exist and leverage it for mobile as well

frett (commented):
the logic looks fine to me, leaving this as just a comment until Justin has a chance to respond to the brakeman policy question

.github/workflows/ruby.yml
jbirdjavi: I think that for this repo, adding that flag should be fine. It's actually more defensible to use it here than it is on mpdx_api. :)

app/controllers/application_controller.rb
```diff
@@ -50,6 +51,13 @@ def authorize!
     render_unauthorized
   end
 
+  # Refresh the user's short-lived, sliding token on every authenticated request.
+  def refresh_auth_token
+    return unless current_user
+
+    response.headers["Authorization"] = AuthToken.new(user: current_user).token
```
frett: is this the header name we use in mpdx_api? (I think that is the other project that does something similar)

canac: I don't think MPDX uses sliding auth tokens. Searching `response.headers` doesn't pull up anything. I'm open to other names if `Authorization` is confusing. Since it's the name of a standard request header.

frett: this comment comes out of the fact that `Authorization` is a request header, it's not a standard response header, and any custom headers should be under an `X-*` name to indicate it's a custom header.

Something like: 
- `X-Fresh-Token`
- `X-Auth-Renewal`

Anyways, if we are already using the `Authorization` response header name in mpdx_api I'd be fine leaving it as is to match our other apps, if not I would prefer it either matches our convention or is a custom header.

frett: maybe it was MissionHub that had the token renewal header, I know it was used somewhere in a mobile app I worked on a few years ago

canac: Yeah, maybe MissionHub: `X-MH-Session` https://github.com/CruGlobal/missionhub-api/blob/635803f7f1510133d0c253dbc0dd9cca09c3d70e/app/controllers/apis/concerns/token_refresh.rb#L13

Do you like `X-MCA-Session`? I think I'd lean towards something more generic like `X-Auth-Renewal`.

frett: yeah, it was missionhub. I just found the android logic for it here: https://github.com/CruGlobal/MissionHub-Android/blob/develop/app/src/main/java/com/missionhub/api/MissionHubApi.java#L422-L428

frett: I'm definitely for something more generic

frett: `X-Auth-Renewal` works for me. It makes the intent clear & avoids any future conflicts on the off chance `Authorization` is added as a response header for some other purpose in the future.

config/environments/development.rb
```diff
@@ -79,7 +79,7 @@
   config.middleware.insert_before 0, Rack::Cors do
     allow do
       origins "*"
-      resource "*", headers: :any, methods: [:get, :post, :put, :patch, :options, :delete]
+      resource "*", headers: :any, methods: [:get, :post, :put, :patch, :options, :delete], expose: ["Authorization"]
```
frett: does this need the name update too?

spec/controllers/application_controller_spec.rb
```diff
@@ -68,6 +68,27 @@ def index
...
+      get :index
+
+      refreshed = response.headers["X-Auth-Renewal"]
+      expect(refreshed).to be_present
+      decoded = AuthToken.decode(refreshed).first
+      expect(decoded["user_id"]).to eq user.id
+      expect(decoded["exp"]).to be_within(5.seconds).of(1.hour.from_now.to_i)
+    end
+
+    it "does not set an Authorization response header when unauthenticated" do
```
frett:
```suggestion
    it "does not set an X-Auth-Renewal response header when unauthenticated" do
```

frett (approved):

